PR #219, opened 2023-01-04
[Quality of life] Added npm tooling script for lint and formatting

uxd-vincent:
this PR does 2 things:
- create scripts "npm run format" and other utilities
- run "npm run format" in the current typescript code and commit the output

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm run format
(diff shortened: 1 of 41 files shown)


uxd-vincent:
> Hello :) It's a good idea to bring tooling to the project!
> 
> Is it possible to have same config for many of our depositories? As example, the app frontend, uses `husky` to add a pre-commit hook that format the code.
> 
> Also, there is `.prettierrc` in the app frontend, may be interesting to have the same one here to have homogenous code.
> 
> Would be great to uniformize: `uxd-client`, `uxd-evm-client` ... as well

for sure, Ima import the same prettierrc

Regarding setting up husky, I think its too early since many lints are failing as of today so I dont know if we want to do that yet

cnek (approved):

PR #221, opened 2023-01-11
[Credix] Add a configurable profits beneficiary

uxd-vincent:
PR Combo:
- uxd-client: https://github.com/UXDProtocol/uxd-client/pull/36
- uxd-program: https://github.com/UXDProtocol/uxd-program/pull/221

PR #214, opened 2022-11-28
[Mercurial] Depository profit collection

GregoryNEUT:

PR #222, opened 2023-01-12
[Mercurial] profits beneficiary

cnek:
much similar to https://github.com/UXDProtocol/uxd-program/pull/221
except: 
- profits_beneficiary_collateral's owner compare directly with depository.profits_beneficiary_key, without passing the profits_beneficiary_key to context
- require non zero pubkey as profits_beneficiary_key when calling collect profit ix, abort otherwise
- added test for above scenraios

client PR: https://github.com/UXDProtocol/uxd-client/pull/37

### DIFF
--- a/programs/uxd/src/error.rs
+++ b/programs/uxd/src/error.rs
@@ -76,7 +76,9 @@ pub enum UxdError {
     InvalidDepositoryShares,
     #[msg("The Profits beneficiary provided does not match the depository's one.")]
     InvalidProfitsBeneficiary,
-    #[msg("The provided mercurial vault does not match the Depository's one.")]
+    #[msg("The Profits beneficiary provided does not match the depository's one.")]
+    ProfitsBeneficiaryNotInitialized,
+    #[msg("The profits beneficiary hasn't setup for this depository")]
     InvalidMercurialVault,
     #[msg("The provided mercurial vault collateral token safe does not match the mercurial vault one.")]
     InvalidMercurialVaultCollateralTokenSafe,
--- a/programs/uxd/src/error.rs
+++ b/programs/uxd/src/error.rs
@@ -74,6 +74,8 @@ pub enum UxdError {
     InvalidDepositoryCollateral,
     #[msg("The provided depository shares does not match the depository's one.")]
     InvalidDepositoryShares,
+    #[msg("The Profits beneficiary provided does not match the depository's one.")]
+    InvalidProfitsBeneficiary,
     #[msg("The provided mercurial vault does not match the Depository's one.")]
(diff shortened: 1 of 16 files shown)


programs/uxd/src/error.rs
```diff
@@ -74,7 +74,11 @@ pub enum UxdError {
     InvalidDepositoryCollateral,
     #[msg("The provided depository shares does not match the depository's one.")]
     InvalidDepositoryShares,
-    #[msg("The provided mercurial vault does not match the Depository's one.")]
+    #[msg("The Profits beneficiary provided does not match the depository's one.")]
+    InvalidProfitsBeneficiary,
+    #[msg("The Profits beneficiary provided does not match the depository's one.")]
+    ProfitsBeneficiaryNotInitialized,
+    #[msg("The profits beneficiary hasn't setup for this depository")]
```
acamill, line 81: The profits beneficiary hasn't setup for this depository
> 
The profits beneficiary hasn't been setup for this depository

programs/uxd/src/events.rs
```diff
@@ -96,6 +96,19 @@ pub struct SetDepositoryMintingDisabledEvent {
...
+#[event]
+pub struct SetDepositoryProfitsBeneficiaryKeyEvent {
+    #[index]
+    pub version: u8,
+    #[index]
+    pub controller: Pubkey,
+    #[index]
+    pub depository: Pubkey,
+    #[index]
+    pub profits_beneficiary_key: Pubkey,
```
acamill, line 109: I wouldn't call that key as we don't use that nomenclature anywhere else

tests/suite/editMercurialVaultDepositorySuite.ts
```diff
@@ -107,13 +107,29 @@ export const editMercurialVaultDepositorySuite = async function ({
       });
     });
 
+    it(`Edit profitsBeneficiaryKey alone should work`, async function () {
```
uxd-vincent, line 110: I applied your comment on my branch, we can simply use: "profitsBeneficiaryCollateral" everywhere instead now :)

programs/uxd/src/error.rs
cnek: grammar 🥲

programs/uxd/src/events.rs
cnek: yes, left the same comments on the crefix PR, was for alignment for the naming,
but gonna get rid of this field now but just `profits_beneficiary_collateral` as discussed

cnek:
aligned much with #221 now, would merge and continue the work on #214

PR #227, opened 2023-01-23
[Proof of concept] [Router] Rebalancing IX for computing depositories targets, add rust integrated testing

uxd-vincent:
Here is my current understanding of the plan:
- Step1: add a "rebalancing weight targets" on each depositories
- Step2: add a "compute rebalancing targets amounts" IX that compute the desired soft target for each depository, permissionless
- Step3: add a "withdraw-everything-over-target-from-credix" permissionless IX that dumps everything into identity depository (not accessible for users)
- Step4: allow users to mint/redeem from identity depository from the frontend
- Step5: Add onchain incentive either through fees or blocking transactions to not mint or redeem past the rebalancing targets

### **This PR aims to implement Step1+Step2 (not step3+ yet)**

TODO: will add CI tests whem the above approach is approved by team


programs/uxd/src/events.rs
```diff
@@ -109,6 +109,19 @@ pub struct SetDepositoryProfitsBeneficiaryCollateralEvent {
     pub profits_beneficiary_collateral: Pubkey,
 }
 
+/// Event called in [instructions::edit_*_depository::handler].
+#[event]
+pub struct SetDepositoryRebalancingTargetWeightEvent {
```
acamill, line 114: I don't think it's rebalancing targets, but simply targetWeight (could be then approached through rebalancing, minting, redeeming etc)

programs/uxd/src/instructions/credix_lp/edit_credix_lp_depository.rs
```diff
@@ -132,6 +134,20 @@ pub(crate) fn handler(
         });
     }
 
+    // optional: rebalancing_target_weight
```
acamill, line 137: Issue with adding it to each depository is that you must have all depositories to validate the value is correct. (It's a relative value)

As such, I recommend pulling it out of the depositories into the controller (as suggested in the Slack message), that way you can modify and make sure the total is still 100%?

programs/uxd/tests/program_uxd/integration_tests/test_credix.rs
```diff
@@ -0,0 +1,84 @@
...
+        program_spl::instructions::process_lamports_airdrop(
+            &mut program_test_context,
+            &keypair.pubkey(),
+            1_000_000_000_000,
+        )
+        .await?;
+    }
+
+    let collateral_mint = Keypair::new();
+    program_spl::instructions::process_token_mint_init(
```
acamill, line 55: Namespace and the multiple indirection might make it confusing, as it feels it's not our domain but it is. But anyway you got the ownership on ship I'm not the main maintainer of the codebase anymore.

Overall good adaptation of tests for the codebase! gg

uxd-vincent: Yeah I understand, but given the massive difficulty of making things work, having a library of function we know just simply work is very helpful instead of having to lookup random repos or undocumented solana functions, just having "known to work instructions" is super helpful

That was the only way to keep my sanity when trying to find what works and what doesnt

uxd-vincent: That being said,  I may definitly have been a bit overboard with it, feel free to let me know if I could reduce the amount of indirection in some places where its excessive

uxd-vincent:
This PR is not needed anymore, all changes here have been merged or will be merged in incremental production-ready PRs

PR #229, opened 2023-02-03
[Cleanup] Removing unused cargo dependencies, and removing fixed floating computations

uxd-vincent:
Hello team,

I spent a lot of time looking at our dependencies in the cargo file while trying to upgrade our anchor version.

Turns out a lot of our dependencies are not needed at all or could be easily removed.

So I made this PR in order to:
- remove all unused dependencies
- remove all floating point computations in favor of simple integer computations, using utility function

This PR will help us in the following way:
- slightly smaller ABI
- less compute used during transaction due to not using floating point computations
- less complicated dependency tree which helps a lot when dealing with depenency conflicts when upgrading anchor


programs/uxd/src/utils/math/compute_amount_after_change.rs
```diff
@@ -0,0 +1,28 @@
+use crate::error::UxdError;
+use anchor_lang::prelude::*;
+
+pub fn compute_amount_after_change(amount_before_change: u128, change: i128) -> Result<u128> {
```
acamill, line 4: Makes things obfuscated I would argue, but we already discussed that and my involvement with this codebase is lower so ok for changes.
I would rename `compute_amount_after_change` to `compute_new_value(value: , delta:)` but nit and prob personal preference.

programs/uxd/src/utils/math/compute_amount_after_change.rs
```diff
@@ -0,0 +1,28 @@
...
+    // In case of a simple positive change (increase), add the two amount
+    if change >= 0 {
+        let increase: u128 = u128::try_from(change).ok().ok_or(UxdError::MathError)?;
+        return Ok(amount_before_change
+            .checked_add(increase)
+            .ok_or(UxdError::MathError)?);
+    }
+    // In case of a negative change, substract using the absolute value (decrease)
+    let decrease: u128 = if change == i128::MIN {
+        // special case: i128::MIN does not have an i128 absolute value
```
acamill, line 14: Has the benefit to cover some edgyedgecase™️

programs/uxd/src/utils/math/compute_amount_after_change.rs
uxd-vincent: yeah naming might not be ideal here indeed, seems like using an `unsigned_accumulator+adding_signed_delta` is a common pattern used a lot in the codebase, so it makes sense having nice utility function instead of inlining a lot of boilerplate imo, but yeah maybe we could stick with unsigned values across the codebase instead, then this function utility would become obsolete ?

uxd-vincent: would the name "compute_amount_plus_delta" be more clear in this case you think ? "compute_new_value" feel a bit vague, and could refer to a lot of differen tthings

programs/uxd/src/utils/math/compute_amount_after_change.rs
uxd-vincent: yeah that's why i made a utility, to make it very very safe to use accross the codebase :)

programs/uxd/src/utils/math/compute_amount_after_change.rs
uxd-vincent: maybe @GregoryNEUT has an opinion about this particular pattern? since you wrote that part of the codebase a while back

programs/uxd/src/utils/math/compute_amount_after_change.rs
cnek: not possible for a token amount change to hit i128::MIN right?
but still nice to have this as this's a general math util fn

programs/uxd/src/state/controller.rs
```diff
@@ -1,6 +1,6 @@
 use crate::error::UxdError;
+use crate::utils::compute_amount_after_change;
 use anchor_lang::prelude::*;
-use fixed::types::I80F48;
 
 pub const MAX_REGISTERED_MANGO_DEPOSITORIES: usize = 8;
```
cnek, line 5: we still get a little bit mango here 🥭 
shouldn't have any usage now

programs/uxd/src/test/utils/math/compute_amount_after_change.rs
```diff
@@ -0,0 +1,95 @@
...
+                }
+                // If the decrease is bigger than the amount, expect failure
+                if decrease > amount_before_change {
+                    prop_assert!(result.is_err());
+                    return Ok(());
+                }
+                // Otherwise we should never fail
+                prop_assert!(result.unwrap() == amount_before_change - decrease);
+                return Ok(());
+            }
```
GregoryNEUT, line 91: No need for `if`. `if change >= 0 {` is true then, the only thing left is `< 0`

```suggestion
            let decrease: u128;
            if change == i128::MIN {
                decrease = u128::try_from(u128::MAX).unwrap() + 1;
            } else {
                decrease = u128::try_from(-change).unwrap();
            }
            // If the decrease is bigger than the amount, expect failure
            if decrease > amount_before_change {
                prop_assert!(result.is_err());
                return Ok(());
            }
            // Otherwise we should never fail
            prop_assert!(result.unwrap() == amount_before_change - decrease);
            Ok(())
```

programs/uxd/src/utils/math/compute_amount_after_change.rs
GregoryNEUT: hm, a lot opinionated. It add an extra layer level, but at same time reduce the boilerplate. Would say get rid of I80F48 is a good thing for now.

Let's go with it for now and see how the codebase evolve with new additions.

For naming idea: `safe_addition`

GregoryNEUT (commented):
Let's go :)

programs/uxd/src/test/utils/math/compute_amount_after_change.rs
uxd-vincent: I agree with this comment, but for some reason rust is having a stroke if I do this???? so confused
Screenshot 2023-02-07 at 5 22 25 PM

uxd-vincent: I agree with this comment, but for some reason rust is having a stroke if I do this???? so confused
Screenshot 2023-02-07 at 5 22 25 PM

programs/uxd/src/utils/math/compute_amount_after_change.rs
acamill: ha yhea right..

acamill (approved):

PR #232, opened 2023-02-21
[WORK IN PROGRESS] [Rebalancing

uxd-vincent:


programs/uxd/tests/integration_tests/api/program_spl/accounts/read_token_account.rs
```diff
@@ -0,0 +1,19 @@
...
+    token_account: &Pubkey,
+) -> Result<spl_token::state::Account, String> {
+    let raw_account = program_test_context
+        .banks_client
+        .get_account(*token_account)
+        .await
+        .map_err(|e| e.to_string())?
+        .ok_or("AccountDoesNotExist")?;
+    if raw_account.owner != spl_token::id() {
+        return Err(String::from("InvalidTokenAccount"));
```
GregoryNEUT, line 16: Could create `error.rs` solely for tests purposes

uxd-vincent: 👌 done in previous PR

programs/uxd/tests/integration_tests/api/program_credix/accounts/create_program_keys.rs
```diff
@@ -0,0 +1,62 @@
+use solana_program::pubkey::Pubkey;
+use solana_sdk::signature::Keypair;
+use solana_sdk::signer::Signer;
+
+const CREDIX_MARKETPLACE_SEED: &str = "this-can-be-whatever";
+
+pub struct ProgramKeys {
```
uxd-vincent, line 7: I was hesitating naming this ProgramInfo everywhere instead of ProgramKeys, but both work, wondering if you guys had any other naming idea

acamill: I think this is confusing.

So there is the concept of ProgramDerivedAddress in Solana, that from input <seeds and a programId> outputs a <pubkey and a bump>. 

I think you should have a helper struct called utils::PDA and inside have methods :

```

pub fn get_lp_token_mint_pda() -> (Pubkey, Bump) {
    program_derived_address("seeds")...
}
``` 

And then you should remove what's not PDA, like authority etc. That would also make this god structure disappear hehe

acamill: And then you can respect dependency injection in each place without stuffing the whole context x)

uxd-vincent: Yeah, tried this before, the problem with that is that not all accounts are PDA and not all accounts can be derived, there is a lot of "keypairs" that you cant simply generate everywhere, you have to pass them throughout the whole thing.

So if we dont have this god structure, we have to pass dozens of parameters everywhere through the whole tree of the tests

uxd-vincent: One thing I could do is to remove the `IdentityDepositoryKeys`, `CredixLpDepositoryKeys`, `MercurialVaultDepositoryKeys `from the `ProgramKeys `(uxd_program) and pass them separately, but not having those structures makes everything very difficult

GregoryNEUT: An other way to handle PDA could be to pass to the function only the information required to find the pdas.

Example, here all you need to find all pda is the `credix_market_seeds`:

```
pub async fn process_create_credix_pass(
    program_test_context: &mut ProgramTestContext,
    credix_market_seeds: String,
    pass_holder: &Pubkey,
    credix_pass: &Pubkey,
    is_investor: bool,
    is_borrower: bool,
    release_timestamp: i64,
    disable_withdrawal_fee: bool,
) -> Result<(), program_test_context::ProgramTestError> {
    let program_state = credix_client::ProgramState::generate_pda().0;
    let global_market_state = credix_client::GlobalMarketState::generate_pda(&market_seeds).0;
    let market_admins = credix_client::MarketAdmins::generate_pda(global_market_state).0;
    
    let accounts = credix_client::accounts::CreateCredixPass {
        owner: program_keys.authority.pubkey(),
        pass_holder: *pass_holder,
        program_state,
        global_market_state,
        credix_pass: *credix_pass,
        market_admins,
        system_program: anchor_lang::system_program::ID,
        rent: anchor_lang::solana_program::sysvar::rent::ID,
    };
    let payload = credix_client::instruction::CreateCredixPass {
        _is_investor: is_investor,
        _is_borrower: is_borrower,
        _release_timestamp: release_timestamp,
        _disable_withdrawal_fee: disable_withdrawal_fee,
    };
    let instruction = Instruction {
        program_id: credix_client::id(),
        accounts: accounts.to_account_metas(None),
        data: payload.data(),
    };
    program_test_context::process_instruction(
        program_test_context,
        instruction,
        &program_keys.authority,
    )
    .await
}
```

programs/uxd/tests/integration_tests/api/program_credix/procedures/process_dummy_actors_behaviors.rs
```diff
@@ -0,0 +1,88 @@
...
+use crate::integration_tests::api::program_credix;
+use crate::integration_tests::api::program_spl;
+use crate::integration_tests::api::program_test_context;
+
+pub async fn process_dummy_actors_behaviors(
+    program_test_context: &mut ProgramTestContext,
+    program_keys: &program_credix::accounts::ProgramKeys,
+    base_token_authority: &Keypair,
+) -> Result<(), program_test_context::ProgramTestError> {
+    // Create a dummy investor
```
GregoryNEUT, line 14: Please remove unecessary comments

programs/uxd/tests/integration_tests/api/program_uxd/instructions/process_edit_credix_lp_depository.rs
```diff
@@ -0,0 +1,125 @@
...
+        assert_eq!(
+            redeemable_amount_under_management_cap_after,
+            redeemable_amount_under_management_cap.unwrap()
+        );
+    } else {
+        assert_eq!(
+            redeemable_amount_under_management_cap_after,
+            redeemable_amount_under_management_cap_before
+        );
+    }
```
GregoryNEUT, line 95: You could simplify the verification by using directly `credix_lp_depository_before` and `credix_lp_depository_after. Not sure why you need to create intermediary variables

```suggestion
    if redeemable_amount_under_management_cap.is_some() {
        assert_eq!(
            credix_lp_depository_after.redeemable_amount_under_management_cap,
            redeemable_amount_under_management_cap.unwrap()
        );
    } else {
        assert_eq!(
            credix_lp_depository_after.redeemable_amount_under_management_cap,
            credix_lp_depository_before.redeemable_amount_under_management_cap
        );
    }
```

GregoryNEUT, line 95: You may use `unwrap_or`:

```suggestion
    assert_eq!(
        redeemable_amount_under_management_cap_after,
        redeemable_amount_under_management_cap
            .unwrap_or(redeemable_amount_under_management_cap_before)
    );
```

GregoryNEUT:
Please run `cargo clippy --all-targets` and fix issues

acamill (approved):

acamill (changes requested):
Approved misclic

programs/uxd/tests/integration_tests/api/program_credix/accounts/create_program_keys.rs
acamill: Yes Greg did develop it well yes!

uxd-vincent: Yes I think both you guys are talking about the same option:
- not passing all data in all cases
- instead pass each parameters one by one only when needed

My point is that:
- this works well for 1st order function (leaf functions)
- but it creates a lot of polution in the high level functions due to the sheer amount of keys/parameters involved in those processes

Seems like you guys are quite interested in trying out this different approach so I will go ahead and rebuild an alternative version of this PR in a separate branch/PR so that we can compare both results

uxd-vincent: @acamill @GregoryNEUT 

I have a 90% polished version of a rebuilt version based on the above feedback here:

https://github.com/UXDProtocol/uxd-program/pull/239

which version do you guys prefer?

uxd-vincent:
> Please run `cargo clippy --all-targets` and fix issues

Addressed in separate PR as this goes outside of this PR's scope: https://github.com/UXDProtocol/uxd-program/pull/240

programs/uxd/tests/integration_tests/api/program_credix/accounts/create_program_keys.rs
uxd-vincent: for posterity, merged alternative approach back into this current branch

programs/uxd/tests/integration_tests/api/program_uxd/accounts/find_accounts.rs
```diff
@@ -0,0 +1,46 @@
+use solana_program::pubkey::Pubkey;
+
+pub fn find_controller() -> Pubkey {
```
acamill, line 3: nit: file could be find_pdas.rs, and funcs could be find_controller_pda (and return the raw output of the func, bumps can be verified in tests etc sometimes)

programs/uxd/tests/integration_tests/api/program_uxd/instructions/process_edit_credix_lp_depository.rs
```diff
@@ -0,0 +1,117 @@
...
+    // Read state before
+    let credix_lp_depository_before =
+        program_test_context::read_account_anchor::<CredixLpDepository>(
+            program_test_context,
+            &credix_lp_depository,
+        )
+        .await?;
+
+    let redeemable_amount_under_management_cap_before =
+        credix_lp_depository_before.redeemable_amount_under_management_cap;
```
acamill, line 42: l42-47 can be in the // THEN part around line 77, all based on l41 `redeemable_amount_under_management_cap_before` and unused till then

acamill (approved):

programs/uxd/tests/integration_tests/api/program_uxd/instructions/process_edit_credix_lp_depository.rs
uxd-vincent: sounds good ser

programs/uxd/tests/integration_tests/api/program_uxd/accounts/find_accounts.rs
uxd-vincent: sounds good ser

uxd-vincent: done

programs/uxd/tests/integration_tests/api/program_uxd/instructions/process_edit_credix_lp_depository.rs
uxd-vincent: done

acamill (approved):

acamill (approved):

PR #237, opened 2023-02-28
[Rebalancing] [Part 0.3] Enforcing strict account sizes

uxd-vincent:
When trying to deserialize an account's data, anchor will throw an error if the size of the datastructure is smaller than the size of the on-chain accounts.

Therefore:
- we need to add the reserved space inside of the accounts data structure so that anchor can parse it
- to harden this process, I created new unit tests and double checked the correctness of the accounts sizes by using the extra explicit notation

### DIFF
--- a/programs/uxd/src/state/controller.rs
+++ b/programs/uxd/src/state/controller.rs
@@ -6,26 +6,27 @@ pub const MAX_REGISTERED_MERCURIAL_VAULT_DEPOSITORIES: usize = 4;
 pub const MAX_REGISTERED_CREDIX_LP_DEPOSITORIES: usize = 4;
 
 // Total should be 885 bytes
+pub const CONTROLLER_RESERVED_SPACE: usize = 230;
 pub const CONTROLLER_SPACE: usize = 8
-    + 1
-    + 1
-    + 1
-    + 32
-    + 32
-    + 1
-    + 255 // Shh. Free real estate
-    + 1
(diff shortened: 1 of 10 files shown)


programs/uxd/src/state/controller.rs
```diff
@@ -74,6 +77,8 @@ pub struct Controller {
     //
     // Total amount of profits collected into the treasury by any depository
     pub profits_total_collected: u128,
+    // For future usage
+    pub _reserved: [u8; CONTROLLER_RESERVED_SPACE],
```
uxd-vincent, line 81: main change for this PR

programs/uxd/src/state/controller.rs
```diff
@@ -6,26 +8,27 @@ pub const MAX_REGISTERED_MERCURIAL_VAULT_DEPOSITORIES: usize = 4;
...
-    + 1
-    + 16
-    + 230;
+    + size_of::<u8>() // bump
+    + size_of::<u8>() // redeemable_mint_bump
+    + size_of::<u8>() // version
+    + size_of::<Pubkey>() // authority
+    + size_of::<Pubkey>() // redeemable_mint
+    + size_of::<u8>() // redeemable_mint_decimals
+    + 255 // _unused, Shh. Free real estate
```
acamill, line 19: The meme lives

acamill (approved):

PR #241, opened 2023-03-08
[Rebalancing] [Part4] Compute depositories target weights IX

uxd-vincent:
PR Combo:
- uxd-client: https://github.com/UXDProtocol/uxd-client/pull/40
- uxd-program: https://github.com/UXDProtocol/uxd-program/pull/241

This PR introduces:
-  a new IX: `compute_depositories_targets`
- 2 new fields for credix/mercurial depositories:
   - `redeemable_amount_under_management_weight` // manual static arbitrary weight desired for the depository
   - `redeemable_amount_under_management_target` // absolute amount depending on the weight, cap and total supply

The goal of this new `compute_depositories_targets` IX the following:
- uses all depositories `redeemable_amount_under_management_weight`
- uses the current total supply and all depositories caps
- then compute the new absolute `redeemable_amount_under_management_target` for all depositories

The idea is that the amount in `redeemable_amount_under_management_target` of each depositories will then be used as an informatory value used during computing rebalancing amounts later (this amount can fluctuate over time, rebalancing can be lagging a bit in time without any problem)


programs/uxd/src/utils/math/compute_amount_fraction.rs
```diff
@@ -0,0 +1,21 @@
+use crate::error::UxdError;
+use crate::utils::checked_convert_u128_to_u64;
+use anchor_lang::prelude::*;
+
+// Precision loss may lower the returned value amount.
+// Precision loss of 1 native unit may be expected.
+pub fn compute_amount_fraction(
```
uxd-vincent, line 7: this is similar to `compute_amount_less_fraction`, but the precision loss is in the oposite direction (which is why we cant use `compute_amount_less_fraction` instead

programs/uxd/tests/integration_tests/api/program_uxd/instructions/process_collect_profits_of_mercurial_vault_depository.rs
```diff
@@ -190,19 +191,23 @@ pub async fn process_collect_profits_of_mercurial_vault_depository(
         )
         .unwrap();
 
-    // Check result
+    // vault_lp_mint.supply must have decreased by the withdrawn profit amount
```
uxd-vincent, line 194: This is an attempt to improve readability by:
- separating chunks of codes
- quick description of what we are looking for checking

uxd-vincent: Based on @acamill feedback on last PR about huge chunks of code being unreadable (which I kinda agree with)

programs/uxd/src/events.rs
```diff
@@ -109,6 +109,19 @@ pub struct SetDepositoryProfitsBeneficiaryCollateralEvent {
...
+/// Event called in [instructions::edit_*_depository::handler].
+#[event]
+pub struct SetDepositoryRedeemableAmountUnderManagementWeightBpsEvent {
+    #[index]
+    pub version: u8,
+    #[index]
+    pub controller: Pubkey,
+    #[index]
+    pub depository: Pubkey,
+    #[index]
```
GregoryNEUT, line 121: I don't think it's useful to index this attribute

programs/uxd/src/instructions/compute_depositories_targets.rs
```diff
@@ -0,0 +1,261 @@
...
+        mut,
+        seeds = [CONTROLLER_NAMESPACE],
+        bump = controller.load()?.bump,
+        constraint = controller.load()?.registered_mercurial_vault_depositories.contains(&mercurial_vault_depository.key()) @UxdError::InvalidDepository,
+        constraint = controller.load()?.registered_credix_lp_depositories.contains(&credix_lp_depository.key()) @UxdError::InvalidDepository,
+    )]
+    pub controller: AccountLoader<'info, Controller>,
+
+    /// #3 The first known active mercurial vault depository
+    #[account(
```
GregoryNEUT, line 31: In the case we create a new depository, this permissionless ix could be called and mess around depositories targets.

We could:
- Use dynamic `remaining_accounts` to pass all registered mercurial vaults and credix vaults depositories.
- Enforce known mercurial/credix depositories `PublicKey`.
- Disable new mercurial/credix depository creation.
- Keep it as it is and know internally that creating a new depository may break rebalancing in a bad way (and don't do that. If we need to add a new repo, need to update the tx first).

programs/uxd/src/instructions/compute_depositories_targets.rs
```diff
@@ -0,0 +1,261 @@
...
+    let credix_lp_depository = &mut ctx.accounts.credix_lp_depository.load_mut()?;
+
+    // ---------------------------------------------------------------------
+    // -- Phase 1
+    // -- Read the desired weights for each depository on chain
+    // -- And generate a raw_target estimations that we can refine later
+    // ---------------------------------------------------------------------
+
+    // We want to balance the supply on weighted portions of the circulating supply for now
+    // This could be based on dynamic on-chain account values and liquidity later
```
GregoryNEUT, line 69: nit: please remove "maybe do later" things. It blurs what you are really doing here, now

programs/uxd/src/instructions/compute_depositories_targets.rs
```diff
@@ -0,0 +1,261 @@
...
+    )?;
+    let credix_lp_depository_raw_target = ctx.accounts.compute_raw_target(
+        redeemable_circulating_supply,
+        credix_lp_depository_weight_bps,
+    )?;
+
+    // ---------------------------------------------------------------------
+    // -- Phase 2
+    // -- Using the raw_target and the depository cap:
+    // -- * Compute the overflow (raw target amount above the cap)
```
GregoryNEUT, line 91: nit: Comments of Phase 1 and Phase 2 use different style ("and" versus bullet points)

programs/uxd/src/instructions/compute_depositories_targets.rs
```diff
@@ -0,0 +1,261 @@
...
+    credix_lp_depository.redeemable_amount_under_management_target_amount =
+        credix_lp_depository_final_target;
+
+    // Success
+    Ok(())
+}
+
+// Into functions
+impl<'info> ComputeDepositoriesTargets<'info> {
+    // Compute a simple raw target: raw_target = total_circulating_supply * weight
```
GregoryNEUT, line 181: nit: comment could be deleted

- Repeating function name in comment have no value.
- Function content is self-explanatory for the formula.

programs/uxd/src/instructions/compute_depositories_targets.rs
```diff
@@ -0,0 +1,261 @@
...
+    mercurial_vault_depository.redeemable_amount_under_management_target_amount =
+        mercurial_vault_depository_final_target;
+    credix_lp_depository.redeemable_amount_under_management_target_amount =
+        credix_lp_depository_final_target;
+
+    // Success
+    Ok(())
+}
+
+// Into functions
```
GregoryNEUT, line 179: nit: outdated

programs/uxd/src/instructions/compute_depositories_targets.rs
uxd-vincent: yeah definitly something we will have to keep in mind if we add new depositories.

Im not sure i understand what you mean by "breaking" the rebalancing.

If we create a new depository, its target will just be 0 until the compute_target IX knows about it, so it will just be automatically allocated to 0%, which kinda make sense if we just add depositories without specifying the rules around it in the code

uxd-vincent:
This PR is now deprecated, another PR is going to be used in its place:
https://github.com/UXDProtocol/uxd-program/pull/249

PR #244, opened 2023-03-15
[Rebalancing] Implement credix withdrawal IXs

uxd-vincent:
This PR introduces 2 new IXs:
 - `rebalance_request_create_from_credix_lp_depository`
     - Create a withdrawal request from credix (during the credix request period)
     - The amount requested is the sum of profits and overflow (over-target amount)
     - Doesnt transfer any funds

 - `rebalance_request_execute_from_credix_lp_depository`
     - Execute a previously created withdrawal request from credix (during the credix redeem period)
     - The amount withdrawn is the sum of profits and overflow (over-target amount)
     - If the amount available is less than the requested amount, the rebalancing is prioritized on best-effort
     - Actually transfer the funds
 
Note: The majority of the diff of this PR is in the integration tests, which help making this change safer


programs/uxd/tests/integration_tests/suites/test_mint_and_redeem.rs
```diff
@@ -202,44 +200,12 @@ async fn test_mint_and_redeem() -> Result<(), program_test_context::ProgramTestE
...
-    let credix_lp_depository = program_uxd::accounts::find_credix_lp_depository_pda(
-        &collateral_mint.pubkey(),
-        &credix_global_market_state,
-    )
-    .0;
-
-    // Set the controller's depositories addresses
-    program_uxd::instructions::process_edit_controller(
+    // Now we set the router depositories to the correct PDAs
+    program_uxd::procedures::process_set_router_depositories(
```
uxd-vincent, line 204: moving this code to this helper function, so that it can be called in other test suites easily

programs/uxd/src/utils/calculate_credix_lp_depository_target_amount.rs
```diff
@@ -0,0 +1,41 @@
...
+            DepositoryInfoForTargetRedeemableAmount {
+                weight_bps: controller.load()?.mercurial_vault_depository_weight_bps,
+                redeemable_amount_under_management_cap: mercurial_vault_depository
+                    .load()?
+                    .redeemable_amount_under_management_cap,
+            },
+        ],
+    )?;
+    Ok(depositories_target_redeemable_amount[0]) // credix is the first in the list
+}
```
uxd-vincent, line 41: I was thinking I could also create a "calculate_all_depositories_target_amount" function instead that return a string and use that for the router mint/redeem too, but I wanted to keep things as small as possible here.

That function definitly feels like a hack tho 🤔

programs/uxd/src/instructions/credix_lp/rebalance_request_create_from_credix_lp_depository.rs
```diff
@@ -0,0 +1,284 @@
...
+    // -- Phase 1
+    // -- Check if the withdraw request period is active at the moment
+    // ---------------------------------------------------------------------
+
+    let start_of_request_phase_timestamp = ctx.accounts.credix_withdraw_epoch.go_live;
+    let end_of_request_phase_timestamp = start_of_request_phase_timestamp
+        .checked_add(ctx.accounts.credix_withdraw_epoch.request_seconds.into())
+        .ok_or(UxdError::MathError)?;
+    let current_unix_timestamp = Clock::get()?.unix_timestamp;
+    if current_unix_timestamp < start_of_request_phase_timestamp
```
acamill, line 159: `require!((start_of_request_phase_timestamp.. end_of_request_phase_timestamp).contains(current_unix_timestamp), Err(UxdError::InvalidCredixWithdrawEpochRequestPeriod.into()));`

nit, should work too

acamill (commented):
Look good to me, better wait additionnal reviews though as it's massive

programs/uxd/src/instructions/credix_lp/rebalance_request_create_from_credix_lp_depository.rs
uxd-vincent: Sounds good! applying now

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
```diff
@@ -0,0 +1,646 @@
...
+    let total_shares_value_before: u64 = liquidity_collateral_amount_before
+        .checked_add(outstanding_collateral_amount_before)
+        .ok_or(UxdError::MathError)?;
+
+    let owned_shares_amount_before: u64 = ctx.accounts.depository_shares.amount;
+    let owned_shares_value_before: u64 = compute_value_for_shares_amount_floor(
+        owned_shares_amount_before,
+        total_shares_supply_before,
+        total_shares_value_before,
+    )?;
```
GregoryNEUT, line 276: nit: scope variables used only locally

```suggestion
    let credix_lp_depository_collateral_amount_before: u64 =
        ctx.accounts.depository_collateral.amount;
    let identity_depository_collateral_amount_before: u64 =
        ctx.accounts.identity_depository_collateral.amount;
    let profits_beneficiary_collateral_amount_before: u64 =
        ctx.accounts.profits_beneficiary_collateral.amount;

    let total_shares_supply_before: u64 = ctx.accounts.credix_shares_mint.supply;
    let total_shares_value_before: u64 = {
        let liquidity_collateral_amount_before: u64 =
            ctx.accounts.credix_liquidity_collateral.amount;
        let outstanding_collateral_amount_before: u64 = ctx
            .accounts
            .credix_global_market_state
            .pool_outstanding_credit;

        liquidity_collateral_amount_before
            .checked_add(outstanding_collateral_amount_before)
            .ok_or(UxdError::MathError)?
    };

    let owned_shares_amount_before: u64 = ctx.accounts.depository_shares.amount;
    let owned_shares_value_before: u64 = compute_value_for_shares_amount_floor(
        owned_shares_amount_before,
        total_shares_supply_before,
        total_shares_value_before,
    )?;
```

programs/uxd/src/instructions/credix_lp/rebalance_request_create_from_credix_lp_depository.rs
```diff
@@ -0,0 +1,283 @@
...
+    // -- Phase 4
+    // -- Emit resulting event
+    // ---------------------------------------------------------------------
+
+    emit!(RebalanceRequestCreateFromCredixLpDepositoryEvent {
+        controller_version: ctx.accounts.controller.load()?.version,
+        depository_version: ctx.accounts.depository.load()?.version,
+        controller: ctx.accounts.controller.key(),
+        depository: ctx.accounts.depository.key(),
+        overflow_value,
```
GregoryNEUT, line 245: would also emit the `requested_collateral_amount`

programs/uxd/src/instructions/credix_lp/rebalance_request_create_from_credix_lp_depository.rs
```diff
@@ -0,0 +1,283 @@
...
+    #[account(mut)]
+    pub credix_withdraw_request: AccountInfo<'info>,
+
+    /// #15
+    pub system_program: Program<'info, System>,
+    /// #16
+    pub credix_program: Program<'info, credix_client::program::Credix>,
+}
+
+pub(crate) fn handler(ctx: Context<RebalanceRequestCreateFromCredixLpDepository>) -> Result<()> {
```
GregoryNEUT, line 139: What happens if we call this instruction twice within the same withdraw request window? Does it overrides? Create a second request?

programs/uxd/src/instructions/credix_lp/rebalance_request_create_from_credix_lp_depository.rs
```diff
@@ -0,0 +1,283 @@
...
+            0
+        } else {
+            redeemable_amount_under_management
+                .checked_sub(redeemable_amount_under_management_target_amount)
+                .ok_or(UxdError::MathError)?
+        }
+    };
+
+    let profits_collateral_amount = {
+        let liquidity_collateral_amount_before: u64 =
```
GregoryNEUT, line 194: nit: no need for all the `before` as we never use the `after` values.

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
```diff
@@ -0,0 +1,646 @@
...
+        .checked_add(ctx.accounts.credix_withdraw_epoch.redeem_seconds.into())
+        .ok_or(UxdError::MathError)?;
+    let end_of_redeem_phase_timestamp = start_of_redeem_phase_timestamp
+        .checked_add(ctx.accounts.credix_withdraw_epoch.redeem_seconds.into())
+        .ok_or(UxdError::MathError)?;
+    require!(
+        (start_of_redeem_phase_timestamp..end_of_redeem_phase_timestamp)
+            .contains(&Clock::get()?.unix_timestamp),
+        UxdError::InvalidCredixWithdrawEpochRequestPeriod,
+    );
```
GregoryNEUT, line 246: nit:

```suggestion
    {
        let start_of_redeem_phase_timestamp = ctx
            .accounts
            .credix_withdraw_epoch
            .go_live
            .checked_add(ctx.accounts.credix_withdraw_epoch.redeem_seconds.into())
            .ok_or(UxdError::MathError)?;
        let end_of_redeem_phase_timestamp = start_of_redeem_phase_timestamp
            .checked_add(ctx.accounts.credix_withdraw_epoch.redeem_seconds.into())
            .ok_or(UxdError::MathError)?;
        require!(
            (start_of_redeem_phase_timestamp..end_of_redeem_phase_timestamp)
                .contains(&Clock::get()?.unix_timestamp),
            UxdError::InvalidCredixWithdrawEpochRequestPeriod,
        );
    }
```

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
```diff
@@ -0,0 +1,646 @@
...
+
+    // All investors gets an equivalent slice of the locked liquidity,
+    // based on their relative position size in the lp pool
+    let withdrawable_total_collateral_amount = locked_liquidity
+        .checked_mul(investor_total_lp_amount)
+        .ok_or(UxdError::MathError)?
+        .checked_div(participating_investors_total_lp_amount)
+        .ok_or(UxdError::MathError)?
+        .checked_sub(base_amount_withdrawn)
+        .ok_or(UxdError::MathError)?;
```
GregoryNEUT, line 338: nit:

```suggestion
    let withdrawable_total_collateral_amount = {
        // Read credix withdrawal accounts onchain state
        let locked_liquidity = ctx.accounts.credix_global_market_state.locked_liquidity;
        let investor_total_lp_amount = ctx
            .accounts
            .credix_withdraw_request
            .investor_total_lp_amount;
        let participating_investors_total_lp_amount = ctx
            .accounts
            .credix_withdraw_epoch
            .participating_investors_total_lp_amount;
        let base_amount_withdrawn = ctx.accounts.credix_withdraw_request.base_amount_withdrawn;

        // All investors gets an equivalent slice of the locked liquidity,
        // based on their relative position size in the lp pool
        locked_liquidity
            .checked_mul(investor_total_lp_amount)
            .ok_or(UxdError::MathError)?
            .checked_div(participating_investors_total_lp_amount)
            .ok_or(UxdError::MathError)?
            .checked_sub(base_amount_withdrawn)
            .ok_or(UxdError::MathError)?
    };
```

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
```diff
@@ -0,0 +1,646 @@
...
+    // ---------------------------------------------------------------------
+
+    let redeemable_amount_under_management = checked_convert_u128_to_u64(
+        ctx.accounts
+            .depository
+            .load()?
+            .redeemable_amount_under_management,
+    )?;
+
+    let profits_collateral_amount = owned_shares_value_before
```
GregoryNEUT, line 290: Here we are recalculating overflow + profit at credix withdraw execution time, which is Xd after the request is made.

Here are two choices:
- Withdraw the exact amount of what have been requested
- Update targets at withdraw time and either withdraw 100% of what have been asked, or <100%.

Note, between the withdraw request time and the execute request time, following things have happened:
- Profits increased
- Possibility of `overflow_value` to be bigger or lower
    - `overflow_value` is bigger now if the UXD circulating supply decreased.
    - `overflow_value` is lower if the UXD circulating supply increased.

GregoryNEUT: Issue I see with updating target is that you may never been able to collect profits as `overflow_value` may increase and grab all available collateral.

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
```diff
@@ -0,0 +1,646 @@
...
+        .ok_or(UxdError::MathError)?
+        .checked_div(participating_investors_total_lp_amount)
+        .ok_or(UxdError::MathError)?
+        .checked_sub(base_amount_withdrawn)
+        .ok_or(UxdError::MathError)?;
+
+    // We prioritize withdrawing the overflow first based on what liquidity is available
+    let withdrawal_overflow_value =
+        std::cmp::min(withdrawable_total_collateral_amount, overflow_value);
+    // Then if liquidity is still available for profits, also withdraw it
```
GregoryNEUT, line 343: Note: we should discuss this one with Kento. Should profit be prioritize versus balance? Especially considering profits represents way smaller number.

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
```diff
@@ -0,0 +1,646 @@
...
+            redeemable_amount_under_management
+                .checked_sub(redeemable_amount_under_management_target_amount)
+                .ok_or(UxdError::MathError)?
+        }
+    };
+
+    // ---------------------------------------------------------------------
+    // -- Phase 4
+    // -- Here we compute how much we would like to withdraw now,
+    // -- We also need to compute how much credix will let us withdraw now,
```
GregoryNEUT, line 314: What if there is not enough when you call this ix? Can you call it again if credix adds more collateral to the pot few hours later and get the rest?

Does this actual ix handle that scenario?

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
```diff
@@ -0,0 +1,646 @@
...
+    )?;
+
+    // Refresh account states after withdrawal
+    ctx.accounts.depository_collateral.reload()?;
+    ctx.accounts.depository_shares.reload()?;
+    ctx.accounts.credix_global_market_state.reload()?;
+    ctx.accounts.credix_liquidity_collateral.reload()?;
+    ctx.accounts.credix_shares_mint.reload()?;
+    ctx.accounts.identity_depository_collateral.reload()?;
+    ctx.accounts.profits_beneficiary_collateral.reload()?;
```
GregoryNEUT, line 429: nit:

```suggestion
    {
        ctx.accounts.depository_collateral.reload()?;
        ctx.accounts.depository_shares.reload()?;
        ctx.accounts.credix_global_market_state.reload()?;
        ctx.accounts.credix_liquidity_collateral.reload()?;
        ctx.accounts.credix_shares_mint.reload()?;
        ctx.accounts.identity_depository_collateral.reload()?;
        ctx.accounts.profits_beneficiary_collateral.reload()?;
    }
```

programs/uxd/src/utils/calculate_credix_lp_depository_target_amount.rs
```diff
@@ -0,0 +1,41 @@
...
+use super::calculate_depositories_target_redeemable_amount;
+
+pub fn calculate_credix_lp_depository_target_amount(
+    controller: &AccountLoader<Controller>,
+    identity_depository: &AccountLoader<IdentityDepository>,
+    mercurial_vault_depository: &AccountLoader<MercurialVaultDepository>,
+    credix_lp_depository: &AccountLoader<CredixLpDepository>,
+) -> Result<u64> {
+    let depositories_target_redeemable_amount = calculate_depositories_target_redeemable_amount(
+        controller.load()?.redeemable_circulating_supply,
```
GregoryNEUT, line 17: Should we call `.load()` once per depository at the start of the function? Not sure what the implications are, more computing?

programs/uxd/src/instructions/credix_lp/rebalance_request_create_from_credix_lp_depository.rs
acamill: ha good one, missed this

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
uxd-vincent: I think I asked this a while back in the slack channel, we can discuss again, but basically the reasoning at the time was:
- an imbalance represent additional risks or insuficient risks
- so an imbalance is more critical than paying out the UXP holders immediately

programs/uxd/src/instructions/credix_lp/rebalance_request_create_from_credix_lp_depository.rs
uxd-vincent: Integration tested over here:
- https://github.com/UXDProtocol/uxd-program/pull/244/files#diff-80dc658a8b356d2727e5b2f9ec2d33568de94bbb63ec3064fc6a6aa5fd9ef953R222

Basically:
- calling 2 create requests in a row will fail, only the first one succeeds
- calling 2 execute requests in a row will succeed, but the amount from the first execution withdrawn is substracted from the second call moved amount

programs/uxd/src/instructions/credix_lp/rebalance_request_create_from_credix_lp_depository.rs
uxd-vincent: Sounds good, changing now!

programs/uxd/src/instructions/credix_lp/rebalance_request_create_from_credix_lp_depository.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    /// #6
+    pub collateral_mint: Box<Account<'info, Mint>>,
+
+    /// #7
+    #[account(mut)]
+    pub depository_shares: Box<Account<'info, TokenAccount>>,
+
+    /// #8
+    #[account()]
```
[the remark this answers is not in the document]
uxd-vincent: Sounds good, changing now!

programs/uxd/src/instructions/credix_lp/rebalance_request_create_from_credix_lp_depository.rs
uxd-vincent: Sounds good, changing now!

programs/uxd/src/utils/calculate_credix_lp_depository_target_amount.rs
uxd-vincent: Yeah why not! applying now!

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
uxd-vincent: Yeah the maths handle this scenario (integration test reflects that in the _illiquid test variant).

Basically the pot is deterministic, based on the maths for the total_withdrawal_amount we know exactly how much we can withdraw now.

And the maths always do compute `min(available, desired)` of all the amounts to account for this

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
uxd-vincent: Yes everything you said is correct. I think we are on the same page.

First of all I think this issue is not a very big problem in practice as the time between request/execute is small and it's a self-healing process as the next month's request will do all the necessary adjustment automatically.

That being said there is something we could do better:
- We could request an "overshot" amount during the create part of the withdraw request
- Then we could execute the exact amount during the execute part of the withdraw request

This would effectively tell credix everymonth "we want to withdraw everything" then we would "actually" just withdraw only what we need and the rest would return to the liquid pool after that if anything we haven't withdrawn

uxd-vincent: I'll make that change now, I don't see why not. That's a good catch! thanks!

Keeping this thread open, let me know if my answer make sense and the proposed solution is sufficient in your opinion

programs/uxd/src/instructions/credix_lp/rebalance_request_create_from_credix_lp_depository.rs
```diff
@@ -0,0 +1,284 @@
...
+use crate::CONTROLLER_NAMESPACE;
+use crate::CREDIX_LP_DEPOSITORY_NAMESPACE;
+use crate::CREDIX_LP_EXTERNAL_PASS_NAMESPACE;
+use crate::CREDIX_LP_EXTERNAL_WITHDRAW_EPOCH_NAMESPACE;
+use crate::IDENTITY_DEPOSITORY_NAMESPACE;
+use crate::MERCURIAL_VAULT_DEPOSITORY_NAMESPACE;
+
+#[derive(Accounts)]
+pub struct RebalanceRequestCreateFromCredixLpDepository<'info> {
+    /// #1 // Permissionless IX that can be called by anyone at any time
```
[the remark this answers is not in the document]
uxd-vincent: commited!

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
uxd-vincent: Commited and applied for the before+after

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
uxd-vincent: Commited!

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
uxd-vincent: Commited!

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
uxd-vincent: Commited! I don't see any local variable here captured in the scope, why did you want to scope this part in particular?

GregoryNEUT: imo scope improve readeability as it link things together. Opiniated

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
uxd-vincent: Discussed by voice chat, team voted to prioritize profits instead of balance. Commited the change!

programs/uxd/src/instructions/credix_lp/rebalance_request_execute_from_credix_lp_depository.rs
uxd-vincent: Discussed more in depth offline, keeping the original maths of just requesting the amount we "think" we are going to be withdrawing during creation phase

acamill (approved):

PR #246, opened 2023-03-28
[Rebalancing] Router Mint IX

uxd-vincent:
This PR introduces a new IX:
 - MintWithRouter
     - which split the collateral minted into each depository based on depositories weights
     - then call mint on all those depositories, sequentially with the computed amounts
     

Example maths values for a scenario of 2 mints in a row (100, then 200):
```python
# First Mint
# Weights set to 10%, 50%, 40%
[mint_with_router]
[mint_with_router:collateral_amount:100000000]
[mint_with_router:identity_depository.redeemable_amount_under_management:0]
[mint_with_router:mercurial_vault_depository_0.redeemable_amount_under_management:0]
[mint_with_router:credix_lp_depository_0.redeemable_amount_under_management:0]
[mint_with_router:depositories_target_redeemable_amount.identity_depository:10000000]
[mint_with_router:depositories_target_redeemable_amount.mercurial_vault_depository_0:50000000]
[mint_with_router:depositories_target_redeemable_amount.credix_lp_depository_0:40000000]
[mint_with_router:depositories_mint_collateral_amount.identity_depository:10000000]
[mint_with_router:depositories_mint_collateral_amount.mercurial_vault_depository_0:50000000]
[mint_with_router:depositories_mint_collateral_amount.credix_lp_depository_0:40000000]

# Second mint
# Weights set to 10%, 40%, 50% (weights have changed since first mint)
[mint_with_router]
[mint_with_router:collateral_amount:200000000]
[mint_with_router:identity_depository.redeemable_amount_under_management:10000000]
[mint_with_router:mercurial_vault_depository_0.redeemable_amount_under_management:50000000]
[mint_with_router:credix_lp_depository_0.redeemable_amount_under_management:39999999] # Note the precision loss
[mint_with_router:depositories_target_redeemable_amount.identity_depository:30000000]
[mint_with_router:depositories_target_redeemable_amount.mercurial_vault_depository_0:120000000]
[mint_with_router:depositories_target_redeemable_amount.credix_lp_depository_0:150000000]
[mint_with_router:depositories_mint_collateral_amount.identity_depository:19999999]
[mint_with_router:depositories_mint_collateral_amount.mercurial_vault_depository_0:69999999]
[mint_with_router:depositories_mint_collateral_amount.credix_lp_depository_0:110000000]
```

PR #250, opened 2023-04-03
[Rebalancing] Router Redeem IX

uxd-vincent:
This PR introduces a new IX:
 - RedeemFromRouter
     - which split the redeemable redeemed from each depository based on depositories weights
     - then call redeem on all those depositories, sequentially with the computed amounts
     

Example maths values for a scenario of 2 mints in a row (20, then 120):
```python
# First Redeem
# Weights set to 0%, 100%, 0% (identity overflowing, mercurial underflowing)
[redeem]
[redeem:redeemable_amount:20000000]
[redeem:identity_depository.redeemable_amount_under_management:29999999]
[redeem:mercurial_vault_depository_0.redeemable_amount_under_management:119999999]
[redeem:credix_lp_depository_0.redeemable_amount_under_management:149999998]
[redeem:depositories_target_redeemable_amount.identity_depository:0]
[redeem:depositories_target_redeemable_amount.mercurial_vault_depository_0:279999996]
[redeem:depositories_target_redeemable_amount.credix_lp_depository_0:0]
[redeem:depositories_redeemable_amount.identity_depository:20000000]
[redeem:depositories_redeemable_amount.mercurial_vault_depository_0:0]
[redeem:depositories_redeemable_amount.credix_lp_depository_0:0]

# Second redeem
# Weights unchanged, mercurial underflowing is still being redeemed from, because no other source available
[redeem]
[redeem:redeemable_amount:120000000]
[redeem:identity_depository.redeemable_amount_under_management:9999999]
[redeem:mercurial_vault_depository_0.redeemable_amount_under_management:119999999]
[redeem:credix_lp_depository_0.redeemable_amount_under_management:149999998]
[redeem:depositories_target_redeemable_amount.identity_depository:0]
[redeem:depositories_target_redeemable_amount.mercurial_vault_depository_0:159999996]
[redeem:depositories_target_redeemable_amount.credix_lp_depository_0:0]
[redeem:depositories_redeemable_amount.identity_depository:9999999]
[redeem:depositories_redeemable_amount.mercurial_vault_depository_0:110000001]
[redeem:depositories_redeemable_amount.credix_lp_depository_0:0]
```

Note: this PR is a follow up to: https://github.com/UXDProtocol/uxd-program/pull/246


programs/uxd/src/instructions/redeem_from_router.rs
```diff
@@ -0,0 +1,350 @@
...
+    #[account(
+        mut,
+        seeds = [CREDIX_LP_DEPOSITORY_NAMESPACE, credix_lp_depository_0.load()?.credix_global_market_state.key().as_ref(), credix_lp_depository_0.load()?.collateral_mint.as_ref()],
+        bump = credix_lp_depository_0.load()?.bump,
+        has_one = controller @UxdError::InvalidController,
+        has_one = collateral_mint @UxdError::InvalidCollateralMint,
+    )]
+    pub credix_lp_depository_0: AccountLoader<'info, CredixLpDepository>,
+
+    /// #23
```
GregoryNEUT, line 135: nit: numbers from here are wrong

programs/uxd/src/instructions/redeem_from_router.rs
```diff
@@ -0,0 +1,350 @@
...
+
+    /// #29
+    pub rent: Sysvar<'info, Rent>,
+}
+
+struct DepositoryInfoForRedeemFromRouter<'info> {
+    pub weight_bps: u16,
+    pub redeemable_amount_under_management: u128,
+    pub redeemable_amount_under_management_cap: u128,
+    pub redeem_cpi: Option<Box<dyn Fn(u64) -> Result<()> + 'info>>,
```
GregoryNEUT, line 158: nit: `redeem_fn`

programs/uxd/src/instructions/redeem_from_router.rs
```diff
@@ -0,0 +1,350 @@
...
+    // Gather all the onchain states we need (caps, weights and supplies)
+    let controller = ctx.accounts.controller.load()?;
+    let identity_depository = ctx.accounts.identity_depository.load()?;
+    let mercurial_vault_depository_0 = ctx.accounts.mercurial_vault_depository_0.load()?;
+    let credix_lp_depository_0 = ctx.accounts.credix_lp_depository_0.load()?;
+
+    // The actual post-redeem circulating supply may be slightly higher
+    // Due to redeem fees and precision loss. But the difference should be negligible and
+    // Any future mint/redeem will recompute the targets based on the exact future circulating supply anyway
+    let mimimum_after_redeem_circulating_supply = controller
```
GregoryNEUT, line 171: At first glance I would say that:

In case of redeem, all the provided UXD are burnt, fees are paid by sending back less collateral to the user.

The calculated circulating supply here is correct.

programs/uxd/src/instructions/redeem_from_router.rs
```diff
@@ -0,0 +1,350 @@
...
+use crate::CONTROLLER_NAMESPACE;
+use crate::CREDIX_LP_DEPOSITORY_NAMESPACE;
+use crate::IDENTITY_DEPOSITORY_COLLATERAL_NAMESPACE;
+use crate::IDENTITY_DEPOSITORY_NAMESPACE;
+use crate::MERCURIAL_VAULT_DEPOSITORY_LP_TOKEN_VAULT_NAMESPACE;
+use crate::MERCURIAL_VAULT_DEPOSITORY_NAMESPACE;
+
+#[derive(Accounts)]
+#[instruction(redeemable_amount: u64)]
+pub struct RedeemFromRouter<'info> {
```
GregoryNEUT, line 27: Nit: `RedeemWithRouter` ? 

We don't exactly redeem from the router, as the router doesn't hold any money. Depositories does.

uxd-vincent: just following the existing naming on this:
- redeem_from_identity_depository
- redeem_from_mercurial_vault_depository

We could name it: `RedeemFromProtocol `if we wanted to make it more gramatically correct?

programs/uxd/src/instructions/redeem_from_router.rs
uxd-vincent: True true, but the precision loss may still impact it and the final circulating supply may be slightly above the computed amount here

programs/uxd/src/instructions/redeem_from_router.rs
uxd-vincent: sounds good

programs/uxd/src/instructions/redeem_from_router.rs
uxd-vincent: nice catch, done!

programs/uxd/src/instructions/redeem_from_router.rs
uxd-vincent: for example:

redeem 7units
- 3units from identity
- 3units from mercurial

total = 6 redeemed vs 7 estimated

programs/uxd/src/instructions/redeem_from_router.rs
GregoryNEUT: just `mint` and `redeem` would work actually.

programs/uxd/src/instructions/redeem_from_router.rs
GregoryNEUT: oh I see, meaning that the user may still have 1 native unit of UXD after redeeming.

should we burn that extra unit to make clean accounting for user? Or put that extra 1 native unit on whatever depository target.

programs/uxd/src/instructions/redeem_from_router.rs
acamill: Yea from/with/to doesn't make much sense in that context, even initially not sure it was the best choice.
Maybe we go with simple Mint and Redeem, without further discrimination (router inferred)

programs/uxd/src/instructions/redeem_from_router.rs
acamill: I think it can be sent to any depository (or identity if you wanna be safe)

uxd-vincent: So actually the maths on this are not trivial for example:
- Lets say you have 1 or 2 native unit of error (from `total-SUM()`)
- You cant just do +1 to identity depository, otherwise if the identity depository is empty or at-limit it make the whole thing fail (since redeeming past 0 is an error)
- You would have to loop over all depository to find one that still has remaining redeemable after the regular redeem
- Since there could be 3 depositories redeemed from, but only one with remaining redeemable space (and it can be a different one than the one with the precision error)

programs/uxd/src/instructions/redeem_from_router.rs
uxd-vincent: Sounds good, changing to mint/redeem in this PR 👌

uxd-vincent: Done!

programs/uxd/src/instructions/redeem_from_router.rs
uxd-vincent: I don't think there is a shortcut that works on all cases here. A counter example would be:

Total Requested 11unit (bump +2 = 13):
 - Depository 1 (25%) -> 13*0.25 = 3
 - Depository 2 (25%) -> 13*0.25 = 3
 - Depository 2 (25%) -> 13*0.25 = 3
 - Depository 2 (25%) -> 13*0.25 = 3
Total Final: 12 - 2 = 10 which is different than initial 11    

Not ideal counter-example, but im pretty sure more such example can be found

uxd-vincent: Yeah that could work, but its probably even more complicated than the actual fix probably, especially the random part. Also I don't know for sure this would work in all cases, as the weights and number of depositories can change 🤔

uxd-vincent: > * You would have to loop over all depository to find one or two that still has remaining redeemable after the regular redeem

I think that would be the easiest safe fix that would work 100%, but unsure if its worth actually doing it ? I can give it a shot

uxd-vincent: Oh wait, but if mercurial has precision loss, we wouldnt be able to account for that anyway 🤔

programs/uxd/src/instructions/redeem.rs
```diff
@@ -0,0 +1,350 @@
...
+use crate::validate_is_program_frozen;
+use crate::CONTROLLER_NAMESPACE;
+use crate::CREDIX_LP_DEPOSITORY_NAMESPACE;
+use crate::IDENTITY_DEPOSITORY_COLLATERAL_NAMESPACE;
+use crate::IDENTITY_DEPOSITORY_NAMESPACE;
+use crate::MERCURIAL_VAULT_DEPOSITORY_LP_TOKEN_VAULT_NAMESPACE;
+use crate::MERCURIAL_VAULT_DEPOSITORY_NAMESPACE;
+
+#[derive(Accounts)]
+#[instruction(redeemable_amount: u64)]
```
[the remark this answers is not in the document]
uxd-vincent: Ah yes good catch! thanks!

programs/uxd/src/instructions/redeem_from_router.rs
GregoryNEUT: What do you mean by that?

Any UXD you give to Mercurial or Credix depository are burned and user receives collateral. Collateral may suffer from precision loss, but it doesn't affect UXD circulating supply/user balance.

uxd-vincent: Ah yes, you're right, struggling a bit to think straight. Let me try to edit `calculate_depositories_redeemable_amount` to try to make it have zero precision loss

uxd-vincent: This should fix this issue here: https://github.com/UXDProtocol/uxd-program/pull/250/commits/a0bd6c43bfdc870cdb9c4b22ae441e20e1b95d28

uxd-vincent: Lets continue the thread over there if we need 👌

acamill (approved):

PR #251, opened 2023-04-05
[Rebalancing] minting collateral amount maths

uxd-vincent:
Since the maths behind computing how much each depository should mint to keep the system balanced is a bit complex:
- I am separating the maths in an utility function
- that way it can be very heavily unit-tested
- Instead of requiring complicated IX testing that would provide less test coverage overall

Note: This PR also fixes a tiny bug in the target amount computation maths where the target may exceed the cap if there is no space at all accross all depositories

### DIFF
--- a/programs/uxd/src/utils/calculate_depositories_mint_collateral_amount.rs
+++ b/programs/uxd/src/utils/calculate_depositories_mint_collateral_amount.rs
@@ -6,7 +6,6 @@ use crate::utils::calculate_depositories_sum_value;
 
 use super::checked_convert_u128_to_u64;
 use super::compute_amount_less_fraction_floor;
-use super::DepositoriesTargetRedeemableAmount;
 
 pub struct DepositoriesMintCollateralAmount {
     pub identity_depository_mint_collateral_amount: u64,
@@ -16,7 +15,9 @@ pub struct DepositoriesMintCollateralAmount {
 
 pub fn calculate_depositories_mint_collateral_amount(
     input_mint_collateral_amount: u64,
-    depositories_target_redeemable_amount: &DepositoriesTargetRedeemableAmount,
+    identity_depository_target_redeemable_amount: u64,
+    mercurial_vault_depository_0_target_redeemable_amount: u64,
+    credix_lp_depository_0_target_redeemable_amount: u64,
     identity_depository_redeemable_amount_under_management: u128,
     mercurial_vault_depository_0_redeemable_amount_under_management: u128,
     credix_lp_depository_0_redeemable_amount_under_management: u128,
@@ -29,18 +30,17 @@ pub fn calculate_depositories_mint_collateral_amount(
     let identity_depository_mintable_collateral_amount =
         calculate_depository_mintable_collateral_amount(
             identity_depository_redeemable_amount_under_management,
-            depositories_target_redeemable_amount.identity_depository_target_redeemable_amount,
+            identity_depository_target_redeemable_amount,
         )?;
     let mercurial_vault_depository_0_mintable_collateral_amount =
         calculate_depository_mintable_collateral_amount(
             mercurial_vault_depository_0_redeemable_amount_under_management,
-            depositories_target_redeemable_amount
-                .mercurial_vault_depository_0_target_redeemable_amount,
+            mercurial_vault_depository_0_target_redeemable_amount,
         )?;
     let credix_lp_depository_0_mintable_collateral_amount =
         calculate_depository_mintable_collateral_amount(
             credix_lp_depository_0_redeemable_amount_under_management,
-            depositories_target_redeemable_amount.credix_lp_depository_0_target_redeemable_amount,
+            credix_lp_depository_0_target_redeemable_amount,
         )?;
 
     // ---------------------------------------------------------------------
--- a/programs/uxd/src/utils/calculate_depositories_mint_collateral_amount.rs
+++ b/programs/uxd/src/utils/calculate_depositories_mint_collateral_amount.rs
@@ -1,0 +1,127 @@
+use anchor_lang::prelude::Result;
+use anchor_lang::require;
+
+use crate::error::UxdError;
+use crate::utils::calculate_depositories_sum_value;
+
+use super::checked_convert_u128_to_u64;
+use super::compute_amount_less_fraction_floor;
+use super::DepositoriesTargetRedeemableAmount;
+
+pub struct DepositoriesMintCollateralAmount {
+    pub identity_depository_mint_collateral_amount: u64,
+    pub mercurial_vault_depository_0_mint_collateral_amount: u64,
+    pub credix_lp_depository_0_mint_collateral_amount: u64,
+}
+
+pub fn calculate_depositories_mint_collateral_amount(
+    total_mint_collateral_amount: u64,
+    depositories_target_redeemable_amount: &DepositoriesTargetRedeemableAmount,
+    identity_depository_redeemable_amount_under_management: u128,
+    mercurial_vault_depository_0_redeemable_amount_under_management: u128,
+    credix_lp_depository_0_redeemable_amount_under_management: u128,
+) -> Result<DepositoriesMintCollateralAmount> {
+    // ---------------------------------------------------------------------
+    // -- Phase 1
+    // -- Calculate the maximum mintable collateral amount for each depository
+    // ---------------------------------------------------------------------
+
+    let identity_depository_mintable_collateral_amount =
+        calculate_depository_mintable_collateral_amount(
+            identity_depository_redeemable_amount_under_management,
+            depositories_target_redeemable_amount.identity_depository_target_redeemable_amount,
+        )?;
+    let mercurial_vault_depository_0_mintable_collateral_amount =
+        calculate_depository_mintable_collateral_amount(
+            mercurial_vault_depository_0_redeemable_amount_under_management,
+            depositories_target_redeemable_amount
+                .mercurial_vault_depository_0_target_redeemable_amount,
+        )?;
+    let credix_lp_depository_0_mintable_collateral_amount =
+        calculate_depository_mintable_collateral_amount(
+            credix_lp_depository_0_redeemable_amount_under_management,
+            depositories_target_redeemable_amount.credix_lp_depository_0_target_redeemable_amount,
+        )?;
+
+    // ---------------------------------------------------------------------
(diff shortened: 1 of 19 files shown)


programs/uxd/src/utils/calculate_depositories_target_redeemable_amount.rs
```diff
@@ -2,38 +2,37 @@ use anchor_lang::prelude::Result;
...
 use super::compute_amount_less_fraction_floor;
 
-pub struct DepositoriesTargets {
-    pub identity_depository_target_amount: u64,
-    pub mercurial_vault_depository_0_target_amount: u64,
-    pub credix_lp_depository_0_target_amount: u64,
+pub struct DepositoriesTargetRedeemableAmount {
+    pub identity_depository_target_redeemable_amount: u64,
+    pub mercurial_vault_depository_0_target_redeemable_amount: u64,
+    pub credix_lp_depository_0_target_redeemable_amount: u64,
```
uxd-vincent, line 14: clarification of the names and units. Since we will have multiple complicated maths functions, this helps differenciate everything

programs/uxd/tests/unit_tests/utils/calculate_depositories_target_redeemable_amount.rs
```diff
@@ -0,0 +1,270 @@
+// Unit tests
+#[cfg(test)]
+mod test_calculate_depositories_target_redeemable_amount {
```
uxd-vincent, line 3: not sure why this appears as an added new file, but this is just the renamed file below, with additional and polished prop test implementation

programs/uxd/src/utils/calculate_depositories_target_redeemable_amount.rs
```diff
@@ -224,32 +229,17 @@ fn calculate_depository_target_amount(
             .checked_sub(depository_available_amount)
             .ok_or(UxdError::MathError)?;
         compute_amount_less_fraction_floor(
-            total_overflow_amount,
+            std::cmp::min(total_overflow_amount, total_available_amount),
```
uxd-vincent, line 232: this is a bugfix surfaced by more advanced prop test

programs/uxd/src/utils/calculate_depositories_sum_value.rs
```diff
@@ -0,0 +1,18 @@
+use anchor_lang::prelude::Result;
+
+use crate::error::UxdError;
+
+/**
+ * Compute the sum of one value for each known depositories
+ */
+pub fn calculate_depositories_sum_value(
```
uxd-vincent, line 8: an utility function that is used everywhere in the rebalancing maths

uxd-vincent: just making an existing function public:
https://github.com/UXDProtocol/uxd-program/pull/251/files#diff-8c17cb1da4f6615510c9e2f8d58fb63cc0a0b776dfa00dfa9e1b586049462dd1L245

GregoryNEUT: If functions becomes utility, we may make it less specific:

```
    pub fn calculate_depositories_sum_value(
        depositories_value: Vec<u64>,
   )
```

programs/uxd/src/utils/calculate_depositories_target_redeemable_amount.rs
```diff
@@ -2,38 +2,37 @@ use anchor_lang::prelude::Result;
...
-    pub mercurial_vault_depository_0_target_amount: u64,
-    pub credix_lp_depository_0_target_amount: u64,
+pub struct DepositoriesTargetRedeemableAmount {
+    pub identity_depository_target_redeemable_amount: u64,
+    pub mercurial_vault_depository_0_target_redeemable_amount: u64,
+    pub credix_lp_depository_0_target_redeemable_amount: u64,
 }
 
-pub fn calculate_depositories_targets(
+pub fn calculate_depositories_target_redeemable_amount(
```
GregoryNEUT, line 17: Same here, we could create a vec to simplify the utils function:

```
pub struct CalculateDepositoriesTargetRedeemableAmountParams {
    weight_bps: u16,
    redeemable_circulating_supply: u128,
    redeemable_amount_under_management_cap: u128,
}

pub fn calculate_depositories_target_redeemable_amount(
    params: Vec<CalculateDepositoriesTargetRedeemableAmountParams>
) -> Result<DepositoriesTargetRedeemableAmount> {
```

uxd-vincent: Actually I wanted to avoid using a vector and made this a utility so that everywhere in the code we know for sure we are taking into account ALL depositories related to rebalancing.

for example if we add a new depository, we just need to update that function's signature and then we get a compilation error everywhere in the code where we need to pass a value

uxd-vincent: Also not using a vector saves on computation since there is no memory allocation involved in just passing parameters from the stack

uxd-vincent: For example here:
https://github.com/UXDProtocol/uxd-program/pull/252/files#diff-9153007ddc0f7276312abc33a288a4535786a530dcc0543818b49d34a18fc988R64

It forces the writer of the code to consider what to do with all depositories, which helps understand the algorithm.

GregoryNEUT: > Actually I wanted to avoid using a vector and made this a utility so that everywhere in the code we know for sure we are taking into account ALL depositories related to rebalancing.
>
> for example if we add a new depository, we just need to update that function's signature and then we get a compilation error everywhere in the code where we need to pass a value.

If I'm adding support for one depository, I would be happy to not have to modify and understand 5 files/funtions. Utility functions are functions you should not have to modify, they are here to be used by bigger-scoped functions to do a simple task.

> Also not using a vector saves on computation since there is no memory allocation involved in just passing parameters from the stack, and since this function is used everywhere, this adds up a lot on the complicated maths IXs

Not sure computation is an issue for the kind of program we are running. Do we run into computation issue as of now?

acamill: Thinking the same as Greg here, gonna be hard to add depositories later and think or what to adapt and what not.

ideally should find a design here when you add a new depository, it errors or fail if not added where it should (either using a check with a total depository count and a vector or some other strat)

programs/uxd/src/utils/calculate_depositories_mint_collateral_amount.rs
```diff
@@ -0,0 +1,127 @@
...
+        )?;
+
+    // ---------------------------------------------------------------------
+    // -- Phase 2
+    // -- Calculate the total amount we could possibly mint
+    // -- If this total is not enough, we abort
+    // ---------------------------------------------------------------------
+
+    let total_mintable_collateral_amount = calculate_depositories_sum_value(
+        identity_depository_mintable_collateral_amount,
```
acamill, line 53: So yes here it should error if we add a new depository and forget to edit i believe else it’s hard to keep track?

programs/uxd/src/utils/calculate_depositories_target_redeemable_amount.rs
uxd-vincent: > If I'm adding support for one depository, I would be happy to not have to modify and understand 5 files/funtions. Utility functions are functions you should not have to modify, they are here to be used by bigger-scoped functions to do a simple task.

True, but in this case in my mind: `the annoyance is a feature not a bug` this is because if you dont consider the files you may silently introduce a bug, which could be very difficult to understand/detect

uxd-vincent: > Thinking the same as Greg here, gonna be hard to add depositories later and think or what to adapt and what not.
> 
> ideally should find a design here when you add a new depository, it errors or fail if not added where it should (either using a check with a total depository count and a vector or some other strat)

Yeah I definitly understand its an overhead, but I think the overhead is trading code development time vs safety:
 - safer because you have to adjust every callsite to check that you did add your new depository
 - less error prone because:
     - the order of parameter is enforced by the naming of the parameters (unlike in a vector where ordering is not guaranteed)
     - the number of parameter can be detected at compile time rather than at run time

uxd-vincent: I think trading:
 - adding a depository quickly is more difficult
 - we are more resilient to typos and tiny errors that are difficult to detect
 
 I think is a good tradeoff to take imo

programs/uxd/src/utils/calculate_depositories_mint_collateral_amount.rs
uxd-vincent: exactly, that was my point, make it compiler-enforced that the algorithm take into account all the necessary depositories

programs/uxd/src/utils/calculate_depositories_target_redeemable_amount.rs
uxd-vincent: I feel you, no worries. I think debates are good overall, but text on PR definitly is not the best format for it.

Maybe in the future we can schedule a 30 minute slot where we can code review by voice and have quick chats to clear up this sort of issues

uxd-vincent: > if one day we add a depository, nothing will remind us of where to modify (no error will appear right? You commented it’s a feature but I don’t think it will warn you? I read on mobile so maybe I miss smthg)

Ah I think i didnt understand this part, yeah nothing is warning us when adding a new depository. But you will be warned if you add the depository to the super-mint and super-redeem, and then it will warn you everywhere in the code further if you do. that's what I meant the feature is

programs/uxd/src/utils/calculate_depositories_target_redeemable_amount.rs
acamill: Nice, can add a comment for the bug maybe

acamill (approved):

uxd-vincent:
Thanks for the review/approval, duly noted for all the feedback.

If I apply the feedbacks it will cancel the approval so to keep the progress momentum going, I will merge as-is and create an iterative PR with the applied feedback.

This will result in a smaller diff to review

PR #252, opened 2023-04-05
[Rebalancing] redeem redeemable amount maths

uxd-vincent:
This PR is a follow up on the minting maths PR: https://github.com/UXDProtocol/uxd-program/pull/251

The goal of this PR is to add an heavily unit-tested utility function that we will use for computing the amount of redeem for a super-redeem IX later on.

Splitting this into a separate PR to keep reviews as safe and easy as possible


programs/uxd/src/utils/calculate_depositories_redeem_redeemable_amount.rs
```diff
@@ -0,0 +1,203 @@
+use anchor_lang::prelude::Result;
+use anchor_lang::require;
+
+use crate::error::UxdError;
+use crate::utils::calculate_depositories_sum_value;
+
+use super::checked_convert_u128_to_u64;
+use super::compute_amount_less_fraction_floor;
+
+pub struct DepositoriesRedeemRedeemableAmount {
```
acamill, line 10: Possibly `DepositoriesRedeemRedeemableAmount` -> `DepositoriesRedeemableAmount`?

programs/uxd/src/utils/calculate_depositories_redeem_redeemable_amount.rs
```diff
@@ -0,0 +1,203 @@
...
+            identity_depository_ideal_redeemable_amount,
+        )?;
+    let mercurial_vault_depository_0_remaining_redeemable_amount =
+        calculate_depository_remaining_redeemable_amount(
+            mercurial_vault_depository_0_redeemable_amount_under_management,
+            mercurial_vault_depository_0_ideal_redeemable_amount,
+        )?;
+    let credix_lp_depository_0_remaining_redeemable_amount = 0; // credix is not liquid
+
+    let total_remaining_redeemable_amount = calculate_depositories_sum_value(
```
acamill, line 64: This code can stay as it is, but since I review I cannot not say what I think x). Not bikeshedding but trying to share my view:

I find there is a lot of indirection and involuntary obfuscation, making things more complicated than they are? Here is why:

You can flatten phases, remove helpers, and make this more straight forward (please correct me if I'm wrong) :
- phase 1 is about: **calculate redeemable_amount_available_above_target**
- phase 2 is about: **calculate redeemable_amount_available_below_target**
- phase 3 is about: **is redeemable_amount_available_above_target enough to fulfil the redeem**
- phase 4 is about: **calculating how much to actually redeem from each depository, weighted to stay on target**


First issue: new terms introduced in the comments and code, where existing term seems to be enough.
 
I believe there are only three concepts here: `amount_to_redeem`, `amount_redeemable_total`, `amount_redeemable_on_target` (and "below target" amount being a derivative), a.k.a "redemption amount requested", "total amount redeemable", "optimal amount redeemable". 
On that particular level, would remove `ideal`/`remaining`/`possible` terms that are all relative, because it's added mental gymnastic:
`Redeemable` somehow reads as "that can possibly be redeemed", it already conveys the idea that's it's about **redeeming** and **possible/remaining**, but what are the "possible" and "ideal" referring too, the cap, the target weight ratio, the total possible.

On the top level I would remove the introduced concept of "phases", here it's "simply" maths to know what's the amount to redeem, and this function is already a maths helper function, that use other helper functions (you see where I'm getting)
   
Second issue: this function is doing work that is unrelated. The results about depository X doesn't infer with the result of depository Y. I would make a single factored function `calculate_redeemable_amount` that takes any depository, and return the amount to redeem, hiding complexity, and then the helpers of the helpers, can be merged in. I have to run but i can write it later to show my point, it's more upgradable (new depository? Just add a `calculate_redeemable_amount(new_depository)` on the top level, which clearly states what it's doing, minimal change instead of growing the current function by 33% size..

acamill: Actually you need all the values to make the calculations

acamill: Would add

```
// new `Redeemable` or `Rebalancable` smthg trait to depositories, with default implementation for 

/// redeemable amount to get to target, 0 if currently underweighted
pub fn redeemable_amount_to_target() -> Result<u64> {  }
```

Then on `Controller` add 

```
/// redeemable amount of r-token protocol wide to get to target, 0 if currently underweighted
pub fn redeemable_amount_to_target(vec<Rebalancable>) -> Result<u64> {  }

/// redeemable amount of r-token protocol wide
pub fn redeemable_amount_available(vec<Rebalancable>) -> Result<u64> {  }
```

Then here you remove that calculation func, and in the main instruction you can have flat straight forward code:

```
pub fn super_redeem_router_whatever() {
 // ...
 let rebalancable_depositories = ...;
 
 require!(rebalancable_depositories, total_registered_depositories); // errors if someone doesn't know about this piece, without assumptions on permissionned/not permissionned (Self contained, doesn't care the rest of the codebase)
 
 let amount_redeemable_total = Controller::redeemable_amount_available(rebalancable_depositories);
 let amount_redeemable_to_target = Controller::redeemable_amount_to_target(rebalancable_depositories);
 
 rebalancable_depositories.map(|depository| 
    // here put your logic of the helper function, in one place, factorised for any depository to come later
    // `calculate_depository_redeem_redeemable_amount `
    // (inside it uses `depository.redeemable_amount_to_target`)
}
```

programs/uxd/tests/unit_tests/utils/calculate_depositories_redeem_redeemable_amount.rs
```diff
@@ -0,0 +1,263 @@
+// Unit tests
+#[cfg(test)]
+mod test_calculate_depositories_redeem_redeemable_amount {
```
acamill, line 3: Nice tests!

programs/uxd/src/utils/calculate_depositories_redeem_redeemable_amount.rs
acamill: Although the current pr is all valid and functionnal/tested, so free to go roll with it, then we get this all merged and can move forward

acamill (approved):

programs/uxd/src/utils/calculate_depositories_redeem_redeemable_amount.rs
uxd-vincent: I named it not just "RedeemableAmount" so that we can know the difference between:
- depositories_target_redeemable_amount
- depositories_redeem_redeemable_amount

the "redeem-redeemable" is a bit redundant, i dont have too much inspiration for this naming, definitly would love to find a better name lol

programs/uxd/src/utils/calculate_depositories_redeem_redeemable_amount.rs
uxd-vincent: No worries, sounds good will do!

uxd-vincent: Good call, this naming is much better

uxd-vincent: I didnt end up needing to extend the controller etc, applying the first comments you mentionned allowed me to minimize the line of code by a lot by just using `vectors.map()` and inlining the logic in lambas in there.

This reads a little bit like a book now, easy to follow reading from top to bottom without jumping

programs/uxd/src/utils/calculate_depositories_redeem_redeemable_amount.rs
acamill: But the redeem is superfluous here I still think, these two variable already have a difference which is the `target`?

depositories_target_redeemable_amount
depositories_redeemable_amount

acamill (approved):

programs/uxd/src/utils/calculate_depositories_redeem_redeemable_amount.rs
uxd-vincent: Sounds good, ok, doing on next PR to not waste this review+approval

PR #260, opened 2023-05-08
[Quality Of Life] removing programId from metadata of uxd.json

uxd-vincent:
Since `anchor build` or `anchor test` constantly auto-delete the meta-data from the json file, this is kind of a pain to having to constantly manually re-add it.

So in this PR I try to just store the programId in the constants file instead of the IDL file directly

### DIFF
--- a/.github/workflows/ci-anchor-test.yml
+++ b/.github/workflows/ci-anchor-test.yml
@@ -204,6 +204,7 @@ jobs:
           sed -i.bak "s/${{ env.TEST_DEV }}/${{ env.TEST_CI_SETUP }}/g" ./Anchor.toml
           npx @project-serum/anchor-cli@${{ env.ANCHOR_VERSION }} test --skip-local-validator --skip-build --skip-deploy
       - name: Returning funded runner wallet SOLs to the bank
+        if: always()
         run: |
           solana transfer --keypair ${{ env.RUNNER_WALLET_PATH }} ${{ env.BANK_WALLET_PATH }} $(echo $(solana balance --keypair ${{ env.RUNNER_WALLET_PATH }} | awk '{print $1}') - ${{ env.TRANSACTION_FEE_OFFSET }} | bc)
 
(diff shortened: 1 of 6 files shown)


tests/constants.ts
```diff
@@ -38,7 +37,7 @@ export const CLUSTER = 'devnet';
 
 // ----------------------------------------------------------------------------
 export const uxdProgramId: PublicKey = new PublicKey(
-  jsonIdl['metadata']['address']
+  'EqMrasrCXsbE6mpXaopUH1qip7auwX4AF5UWWpXaLr6u'
```
cnek, line 40: need to take extra care on checking whether this program id matched the one on `./target/deploy/uxd-keypair.json`, but let's go for this changes coz the removal of program id with `anchor build` is indeed super annoying

uxd-vincent: For sure, True!

Before we had 3 places to double check: lib.rs, uxd.json, Anchor.toml

Now we have 3 places to double check: lib.rs, constants.ts, Anchor.toml

No real improvements here, just avoiding the auto-generation problem that nukes uxd.json

.github/workflows/ci-anchor-test.yml
```diff
@@ -120,6 +120,7 @@ jobs:
           npx @project-serum/anchor-cli@${{ env.ANCHOR_VERSION }} build
           solana program deploy ./target/deploy/uxd.so --program-id ${{ env.RESIDENT_PROGRAM_KEYPAIR }} --upgrade-authority ${{ env.CI_DEPLOYMENT_AUTHORITY_WALLET_PATH }}
       - name: Returning funded runner wallet SOLs to the bank
+        if: always()
```
cnek, line 123: thx for saving our precious SOL

cnek (approved):

uxd-vincent:
Getting those errors:

> Error: Metadata property not found in IDL of program: uxd
inside of the Setup Uxd Suite job

It looks like this change might not be possible after all, closing for now, this is not urgent

PR #263, opened 2023-05-19
[Rebalancing] Optional small cleanups

uxd-vincent:


programs/uxd/src/instructions/mint.rs
```diff
@@ -281,44 +291,22 @@ pub(crate) fn handler(ctx: Context<Mint>, collateral_amount: u64) -> Result<()>
         },
     ];
 
-    drop(controller);
-    drop(identity_depository);
-    drop(mercurial_vault_depository);
-    drop(credix_lp_depository);
-
-    // Compute the desired target amounts for each depository
-    let depositories_target_redeemable_amount = calculate_depositories_target_redeemable_amount(
```
uxd-vincent, line 290: this logic is simplified here, that was the goal of re-using the target computing function

uxd-vincent:
not doing this for now

PR #271, opened 2023-06-26
[Tooling] Using credix pre-mainnet instead of devnet for CI

uxd-vincent:
Moving to latest credix testing program: pre-mainnet
Also removing tests for `collectProfitsFromCredix` and `redeemFromCredix` since those two IX are now replaced by the rebalancing logic instead and wont work in the latest mainnet


programs/uxd/tests/integration_tests/api/program_credix/binaries/dump.sh
```diff
@@ -1,5 +1,5 @@
 solana config set --url https://api.devnet.solana.com
-solana program dump crdszSnZQu7j36KfsMJ4VEmMUTJgrNYXwoPVHUANpAu executable-devnet.so
+solana program dump crdRi38zEhQdzpsxnKur73WHBM9BSvXMSfGcbLyJCdP executable-devnet.so
```
cnek, line 2: call it executable-pre-mainnet?

package.json
```diff
@@ -4,7 +4,7 @@
     "@project-serum/anchor": "0.26.0",
     "@solana/spl-token": "0.3.7",
     "@solana/web3.js": "1.73.0",
-    "@uxd-protocol/uxd-client": "7.3.1-rc1"
+    "@uxd-protocol/uxd-client": "7.2.0-rc2"
```
cnek, line 7: should it stay `7.3.1-rc1` as discussed?

programs/uxd/tests/integration_tests/api/program_credix/binaries/dump.sh
uxd-vincent: Well, it does run on the devnet tho T_T

package.json
uxd-vincent: for sure, will revert 👌

programs/uxd/tests/integration_tests/api/program_credix/binaries/dump.sh
cnek: yea let's keep the name then coz it's the only legit devnet program we should use going forward

acamill (dismissed):

cnek (approved):

PR #280, opened 2023-07-21
Scaffold LSD Depository

acamill:

PR #283, opened 2023-07-28
Math refactor

cnek:
breaking down #280 
the renaming part of existing ixs,
- MathError -> MathOverflow
- BPS_POW -> BPS_DECIMAL
- BPS_UNIT_CONVERSION ->  BPS_POWER


.vscode/settings.json
```diff
@@ -69,5 +69,6 @@
     "wsol",
     "WSOLATA"
   ],
-  "editor.formatOnSave": true
+  "editor.formatOnSave": true,
+  "rust-analyzer.linkedProjects": ["./programs/uxd/Cargo.toml"]
```
uxd-vincent, line 73: nit: i think we can remove this?

programs/uxd/src/utils/maths.rs
```diff
@@ -0,0 +1,406 @@
...
+        )?)
+    } else {
+        checked_as_u64(checked_div(
+            checked_div(scaled_coeff1, coefficient2 as u128)?,
+            checked_pow(10u128, (-target_power) as usize)?,
+        )?)
+    }
+}
+
+pub fn checked_decimal_ceil_div(
```
uxd-vincent, line 132: some of those i dont see us using it a lot, is that a lib imported from somewhere ? or is that functions used for the future ? do we want to write unit tests for some of those ?

uxd-vincent:
LGTM, just wondering where those functions come from ? most of the new functions look unused, do we want to commit only the ones we need ? or is that part of a copy/pasted lib ?

uxd-vincent (dismissed):

.vscode/settings.json
acamill: Yhea not sure what's used for, we can

programs/uxd/src/utils/maths.rs
acamill: It's all used by several solana labs projects, audited several times, I kept the whole thing (I think it's better to use that and go there when we need something than rewrite. Would keep as it and update when they do from time to time)

acamill:
Ideally we would replace the old maths with the new one, but let's start with new code instead

acamill (approved):

.vscode/settings.json
uxd-vincent: I'm not sure what you mean, what commands are you running ?

If this is useful we can keep it tho, not a big deal!

uxd-vincent (approved):

acamill (approved):

PR #288, opened 2023-08-02
[Router] Enforce redeem outflow limitations

uxd-vincent:
This is an implementation for the limiting the outflows to a hardcoded amount (or BPS) per epoch (epoch time configurable)

Basically maintain a set of new controller flags, and adds some accounting in the mint/redeem router IXs

PR Combo:
- uxd-client: https://github.com/UXDProtocol/uxd-client/pull/52
- uxd-cpi: https://github.com/UXDProtocol/uxd-cpi/pull/8
- uxd-program: https://github.com/UXDProtocol/uxd-program/pull/288

TLDR of the logic basically:
- when you redeem, increase the `epoch_outflow_amount`
- when you redeem, if you `epoch_outflow_amount` too much, stop the redeem
- when you mint, decrease the `epoch_outflow_amount`
- when you redeem, if some time has passed since the last redeem, we allow for more redeem based on how long was elapsed (decreased the `epoch_outflow_amount` proportionally to how long was elapsed since last redeem)


programs/uxd/src/instructions/redeem.rs
```diff
@@ -160,16 +163,66 @@ struct DepositoryInfoForRedeem<'info> {
...
+            .ok_or(UxdError::MathError)?
+    } else {
+        0
+    };
+
+    // Make sure we are not over the outflow limit!
+    require!(
+        u128::from(last_day_outflow_amount) <= limit_outflow_amount_per_day,
+        UxdError::MaximumOutflowAmountError
+    );
```
uxd-vincent, line 218: Main logic is here

programs/uxd/src/instructions/mint.rs
```diff
@@ -203,16 +203,23 @@ struct DepositoryInfoForMint<'info> {
 
 pub(crate) fn handler(ctx: Context<Mint>, collateral_amount: u64) -> Result<()> {
     // Gather all the onchain states we need (caps, weights and supplies)
-    let controller = ctx.accounts.controller.load()?;
+    let mut controller = ctx.accounts.controller.load_mut()?;
     let identity_depository = ctx.accounts.identity_depository.load()?;
     let mercurial_vault_depository = ctx.accounts.mercurial_vault_depository.load()?;
     let credix_lp_depository = ctx.accounts.credix_lp_depository.load()?;
 
+    // Make sure minting reduces the counter of outflow (minting is inflow)
```
acamill, line 211: 👍

target/idl/uxd.json
```diff
@@ -3398,136 +3436,141 @@
     },
     {
       "code": 6030,
```
acamill, line 3438: For this reason I think we should append error at the end of the file (so that error codes keep the same meaning for the future when someone look in the past tx onchain)

programs/uxd/src/state/controller.rs
```diff
@@ -94,6 +97,11 @@ pub struct Controller {
     pub mercurial_vault_depository: Pubkey,
     pub credix_lp_depository: Pubkey,
 
+    // Redeem limitation flags
+    pub limit_outflow_amount_per_day: u64, // or limit_outflow_bps_per_day
```
acamill, line 101: If by `limit_outflow_bps_per_day` you mean that it become something proportionnal to the current total circulating supply then yes I think it needs to be (like 5% or smthg). Then come the zeno paradox when the protocol is fully redeemed though... but minor I think? And we can put a hardcodecd `max(10k or proportionnal value)`

programs/uxd/src/instructions/redeem.rs
```diff
@@ -160,16 +163,66 @@ struct DepositoryInfoForRedeem<'info> {
...
+
+    // How much was unlocked by waiting since last redeem
+    let unlocked_outflow_amount = checked_convert_u128_to_u64({
+        u128::try_from(
+            now_timestamp
+                .checked_sub(controller.last_redeem_timestamp)
+                .ok_or(UxdError::MathError)?,
+        )
+        .ok()
+        .ok_or(UxdError::MathError)?
```
acamill, line 192: Could we base this PR off the Math PR?

acamill: I think all new code should use the new maths to not create more refactor for later, if you'r okay with the math PR

programs/uxd/src/instructions/redeem.rs
```diff
@@ -160,16 +163,66 @@ struct DepositoryInfoForRedeem<'info> {
...
 pub(crate) fn handler(ctx: Context<Redeem>, redeemable_amount: u64) -> Result<()> {
     // Gather all the onchain states we need (caps, weights and supplies)
-    let controller = ctx.accounts.controller.load()?;
+    let mut controller = ctx.accounts.controller.load_mut()?;
     let identity_depository = ctx.accounts.identity_depository.load()?;
     let mercurial_vault_depository = ctx.accounts.mercurial_vault_depository.load()?;
     let credix_lp_depository = ctx.accounts.credix_lp_depository.load()?;
+    let now_timestamp = Clock::get()?.unix_timestamp;
+
+    // Limit per day (NO MERGE, this include both options!! we only need one!!)
```
acamill, line 172: I think for the Zeno paradox thing mentioned earlier... We need both? And we can keep both parameterised, although need proper commenting to explain why

programs/uxd/src/instructions/redeem.rs
```diff
@@ -160,16 +163,66 @@ struct DepositoryInfoForRedeem<'info> {
...
+    } else {
+        controller
+            .redeemable_circulating_supply
+            .checked_mul(1000 /* controller.limit_outflow_bps_per_day */)
+            .ok_or(UxdError::MathError)?
+            .checked_div(BPS_UNIT_CONVERSION.into())
+            .ok_or(UxdError::MathError)?
+    };
+
+    // How much was unlocked by waiting since last redeem
```
acamill, line 184: ** I think the comment is confusing. Maybe worth removing the concept of days, introducing a "epoch" concept (which can be any slots duration, a day can be the starting value). Then this can become an uncommented clearer named var like `epoch_outflow_amount`

(I'll tag the related comments with the starting **)

programs/uxd/src/instructions/redeem.rs
```diff
@@ -160,16 +163,66 @@ struct DepositoryInfoForRedeem<'info> {
...
+        .ok()
+        .ok_or(UxdError::MathError)?
+        .checked_mul(limit_outflow_amount_per_day)
+        .ok_or(UxdError::MathError)?
+        .checked_div(SECONDS_PER_DAY.into())
+        .ok_or(UxdError::MathError)?
+    })?;
+
+    // How much outflow after this current redeem IX
+    let after_redeem_daily_outflow_amount = controller
```
acamill, line 200: ** `new_epoch_outflow_amount`

programs/uxd/src/instructions/redeem.rs
```diff
@@ -160,16 +163,66 @@ struct DepositoryInfoForRedeem<'info> {
...
+        .last_day_outflow_amount
+        .checked_add(redeemable_amount)
+        .ok_or(UxdError::MathError)?;
+
+    // How much outflow after this and since last redeem
+    let last_day_outflow_amount = if after_redeem_daily_outflow_amount > unlocked_outflow_amount {
```
acamill, line 206: ** `previous_epoch_outflow_amount`

programs/uxd/src/instructions/redeem.rs
acamill: Good pr overall, thanks for taking a jab at it ser

programs/uxd/src/state/controller.rs
acamill, line 101: ** `outflow_limit_per_epoch`

programs/uxd/src/state/controller.rs
```diff
@@ -94,6 +97,11 @@ pub struct Controller {
...
+    pub last_day_outflow_amount: u64,
```
acamill, line 102: ** `last_epoch_outflow_amount`

programs/uxd/src/state/controller.rs
```diff
@@ -94,6 +97,11 @@ pub struct Controller {
...
+    pub last_redeem_timestamp: i64,
```
acamill, line 103: ** `last_outflow_timestamp` ?

programs/uxd/src/state/controller.rs
uxd-vincent: Yeah for sure king, agreed, I think I will provide support for both 👌 to avoid the infinite redeem tail

target/idl/uxd.json
uxd-vincent: Smart, ok will do 👌

programs/uxd/src/instructions/redeem.rs
uxd-vincent: For sure, added support for both BPS and amount (and doing the max of the two)

This should solve this issue and let us be quite configurable here

programs/uxd/src/instructions/redeem.rs
uxd-vincent: Sounds good 👌

programs/uxd/src/instructions/redeem.rs
uxd-vincent: Sounds good 👌

programs/uxd/src/state/controller.rs
uxd-vincent: Sounds good 👌

programs/uxd/src/state/controller.rs
uxd-vincent: Sounds good 👌

programs/uxd/src/state/controller.rs
uxd-vincent: Sounds good 👌

programs/uxd/src/instructions/redeem.rs
uxd-vincent: Sounds good 👌 I changed the names to epoch and added a "seconds_per_epoch" configuration

programs/uxd/src/instructions/redeem.rs
uxd-vincent: Done 👌 

Merged math PR into this one 👌

uxd-vincent: Also used the new functions 🥳

programs/uxd/src/instructions/edit_controller.rs
```diff
@@ -139,12 +186,13 @@ impl<'info> EditController<'info> {
 
         // Validate the depositories_routing_weight_bps if specified
         if let Some(depositories_routing_weight_bps) = fields.depositories_routing_weight_bps {
-            let total_weight_bps = depositories_routing_weight_bps
-                .identity_depository_weight_bps
-                .checked_add(depositories_routing_weight_bps.mercurial_vault_depository_weight_bps)
-                .ok_or(UxdError::MathOverflow)?
-                .checked_add(depositories_routing_weight_bps.credix_lp_depository_weight_bps)
-                .ok_or(UxdError::MathOverflow)?;
+            let total_weight_bps = checked_add(
```
acamill, line 189: any checked_sum in existance? prob not but 😆

uxd-vincent: Didnt find any :p

programs/uxd/src/instructions/mint.rs
```diff
@@ -203,24 +205,28 @@ struct DepositoryInfoForMint<'info> {
...
+    controller.epoch_outflow_amount = if controller.epoch_outflow_amount > collateral_amount {
```
acamill, line 214: Maybe this is more explicit:

```suggestion
    controller.epoch_outflow_amount = min(0, checked_sub(controller.epoch_outflow_amount, collateral_amount)?);
```

acamill: Sorry I don't know how to use suggestion properly

uxd-vincent: We could make a math function for this :D

programs/uxd/src/instructions/redeem.rs
```diff
@@ -160,24 +167,67 @@ struct DepositoryInfoForRedeem<'info> {
...
+            checked_as_u128(last_outflow_seconds_ago)?,
+            current_outflow_limit_per_epoch_amount,
+        )?,
+        controller.seconds_per_epoch.into(),
+    )?)?;
+
+    // How much outflow in the current epoch right before the redeem IX
+    let previous_epoch_outflow_amount = if controller.epoch_outflow_amount > unlocked_outflow_amount
+    {
+        checked_sub(controller.epoch_outflow_amount, unlocked_outflow_amount)?
```
acamill, line 205: same

programs/uxd/src/instructions/mint.rs
acamill: Ok no math function we'r good I take it back x)

programs/uxd/src/instructions/redeem.rs
uxd-vincent: I'm confused, I think I should leae it as-is right ? based on your comment above: https://github.com/UXDProtocol/uxd-program/pull/288#discussion_r1285332840

programs/uxd/src/state/controller.rs
```diff
@@ -94,6 +101,13 @@ pub struct Controller {
     pub mercurial_vault_depository: Pubkey,
     pub credix_lp_depository: Pubkey,
 
+    // Flags needed for outflow limitation
+    pub outflow_limit_per_epoch_amount: u64,
+    pub outflow_limit_per_epoch_bps: u16,
+    pub seconds_per_epoch: u32,
```
acamill, line 107: Not sure should it be seconds or slots, prob slots?

programs/uxd/src/state/controller.rs
```diff
@@ -94,6 +101,13 @@ pub struct Controller {
...
+    pub epoch_outflow_amount: u64,
+    pub last_outflow_timestamp: i64,
```
acamill, line 109: Maybe `epoch_start_timestamp`?

acamill (approved):
Looking good, have made some comments

programs/uxd/src/state/controller.rs
uxd-vincent: Looking into the comments, I double checked and I think this name make semantic sense:
`last_outflow_timestamp` is the timestamp of the last redeem, it is not the start of the epoch (the epoch start is constantly X seconds ago, we dont need to store that I believe)

We just need this flag to compute "how long ago was the last redeem" so that we know how much we can decrease the current `epoch_outflow_amount` by.

Maybe i'm miss understanding your comment tho 👌 please let me know

GregoryNEUT (commented):
Looking good :) 

Note: got confused at first because of `epoch` terminology that is similar to solana epochs

programs/uxd/src/instructions/redeem.rs
```diff
@@ -160,24 +167,69 @@ struct DepositoryInfoForRedeem<'info> {
...
+    let controller_pda_signer: &[&[&[u8]]] = &[&[CONTROLLER_NAMESPACE, &[controller.bump]]];
 
     // The actual post-redeem circulating supply may be slightly higher
     // Due to redeem fees and precision loss. But the difference should be negligible and
     // Any future mint/redeem will recompute the targets based on the exact future circulating supply anyway
-    let mimimum_after_redeem_circulating_supply = controller
-        .redeemable_circulating_supply
-        .checked_sub(redeemable_amount.into())
-        .ok_or(UxdError::MathOverflow)?;
+    let mimimum_after_redeem_circulating_supply = checked_sub(
```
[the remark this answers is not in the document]
uxd-vincent: nice catch haha

programs/uxd/tests/integration_tests/suites/test_mint_and_redeem.rs
```diff
@@ -298,10 +340,54 @@ async fn test_mint_and_redeem() -> Result<(), program_test_context::ProgramTestE
...
+            depositories_routing_weight_bps: None,
+            router_depositories: None,
+            outflow_limit_per_epoch_amount: Some(amount_for_second_redeem),
+            outflow_limit_per_epoch_bps: None,
+            seconds_per_epoch: None,
+        },
+    )
+    .await?;
+
     // Redeeming after we exhaused the identity depository should fallback to mercurial depository
```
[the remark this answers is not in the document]
uxd-vincent: nice catch!

programs/uxd/src/instructions/redeem.rs
```diff
@@ -160,24 +167,69 @@ struct DepositoryInfoForRedeem<'info> {
...
+    let unlocked_outflow_amount = checked_as_u64(checked_div(
+        checked_mul(
+            checked_as_u128(last_outflow_seconds_ago)?,
+            checked_as_u128(outflow_limit_per_epoch_amount)?,
+        )?,
+        checked_as_u128(controller.seconds_per_epoch)?,
+    )?)?;
+
+    // How much outflow in the current epoch right before the redeem IX
+    let previous_epoch_outflow_amount = if controller.epoch_outflow_amount > unlocked_outflow_amount
```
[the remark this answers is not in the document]
uxd-vincent: oh I love this

programs/uxd/src/instructions/redeem.rs
```diff
@@ -160,24 +167,69 @@ struct DepositoryInfoForRedeem<'info> {
...
     let mercurial_vault_depository = ctx.accounts.mercurial_vault_depository.load()?;
     let credix_lp_depository = ctx.accounts.credix_lp_depository.load()?;
 
+    // Compute real outflow limit for this epoch (max of bps/amount options)
+    // Note: intermediary maths forced to use u128 to be able to multiply u64s safely
+    let outflow_limit_per_epoch_amount = std::cmp::max(
+        controller.outflow_limit_per_epoch_amount,
+        checked_as_u64(checked_div(
+            checked_mul(
+                checked_as_u128(controller.redeemable_circulating_supply)?,
```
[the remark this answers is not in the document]
uxd-vincent: true! I still want to keep the enforcement here because: if we change the type to u64 one day, this might accidentally make this multiplication become u64*u64 which would be a problem, I prefer forcing everything onto u128 explicitely, and if one conversion is a no-op its no big deal IMO :)

I just like the explicitness about it: if we RELY on the behavior of u128, i prefer putting that in the code rather than letting the type inference figure it out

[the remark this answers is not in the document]
uxd-vincent: ah sad, this throws a lint warning, I guess i'll just remove this for now

programs/uxd/src/state/controller.rs
uxd-vincent: Sounds good 👌  changed to slots_per_epoch and changed other places in the code to use the sysvar::Clock.slot instead

programs/uxd/src/instructions/redeem.rs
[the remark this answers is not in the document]
GregoryNEUT: If you want to enforce it for calculations, make `u128` explicit, such as:

```
checked_mul::<u128>(
      controller.redeemable_circulating_supply,
      checked_as_u128(controller.outflow_limit_per_epoch_bps)?,
 )?,
```

programs/uxd/src/instructions/redeem.rs
```diff
@@ -149,6 +155,9 @@ pub struct Redeem<'info> {
 
     /// #21
     pub rent: Sysvar<'info, Rent>,
+
+    /// #22
+    pub clock: Sysvar<'info, Clock>,
```
cnek, line 160: clock could be accessed using `Clock::get()` I think, so not necessary to pass the this account

programs/uxd/src/instructions/edit_controller.rs
```diff
@@ -121,6 +128,46 @@ pub(crate) fn handler(ctx: Context<EditController>, fields: &EditControllerField
...
+        });
+    }
+
+    // Optionally edit "outflow_limit_per_epoch_bps"
+    if let Some(outflow_limit_per_epoch_bps) = fields.outflow_limit_per_epoch_bps {
+        msg!(
+            "[edit_controller] outflow_limit_per_epoch_bps {}",
+            outflow_limit_per_epoch_bps
+        );
+        controller.outflow_limit_per_epoch_bps = outflow_limit_per_epoch_bps;
```
cnek, line 152: better with a extra checking to ensure the input bps <= BPS_POWER

programs/uxd/src/instructions/mint.rs
```diff
@@ -203,24 +204,26 @@ struct DepositoryInfoForMint<'info> {
...
-    let controller = ctx.accounts.controller.load()?;
+    let mut controller = ctx.accounts.controller.load_mut()?;
     let identity_depository = ctx.accounts.identity_depository.load()?;
     let mercurial_vault_depository = ctx.accounts.mercurial_vault_depository.load()?;
     let credix_lp_depository = ctx.accounts.credix_lp_depository.load()?;
 
+    // Make sure minting reduces the counter of outflow (minting is inflow)
+    controller.epoch_outflow_amount = controller
+        .epoch_outflow_amount
+        .saturating_sub(collateral_amount);
```
cnek, line 215: um...if there are more inflow than outflow momentarily in this epoch, should this counter goes negative i.e. `epoch_outflow_amount` is of type i64/i128?

uxd-vincent: so actually this is using "saturating_sub" which does not go negative for u64 :) I believe this should just be capped to zero in that case

programs/uxd/src/instructions/edit_controller.rs
uxd-vincent: Good point ser! doing now 👌

acamill (approved):

PR #289, opened 2023-08-06
[Credix] Bug Fix: Under-requested rebalancing

uxd-vincent:
Investigating a rebalancing error:
https://explorer.solana.com/tx/4Ui2dkt1RHKRbXQfq6yS5zqEyFCzJp7dQg2RbRZHEjyte7jmnyAeXP8iD6ggEr7HHmUCJpZPXLspTH9DiCKtqjgV

Basically the following scenario would happen sometimes:
- 1) We create the withdraw request (rebalancing request), we compute the amount withdrawable based on the profits at this time
- 2) Right after we create the withdraw request (and before the withdraw redeem) new profits arrive
- 3) The Withdraw Redeem is then triggered, but it is trying to withdraw more profits than we requested in step 1 (since profit increased in the meantime)
- 4) If (and only if) the liquidity is not enough for step3's profit, and someone else is also trying to withdraw that liquidity, credix is capping the redeem to the amount requested, making the rebalance fail

This PR fixes this:
- 1) create a reproduction unit test
- 2) take the request amount into account when doing the redeem
- 3) cleanup the maths wherever possible and wherever i'm testing


programs/uxd/src/instructions/credix_lp/rebalance_redeem_withdraw_request_from_credix_lp_depository.rs
```diff
@@ -329,19 +336,23 @@ pub(crate) fn handler(
...
+            )?,
+            u128::from(participating_investors_total_lp_amount),
+        )?)?;
+        // Finished, just need to take into account the amount already withdrawn and the requested amount
+        let requested_base_amount = ctx.accounts.credix_withdraw_request.base_amount;
+        let withdrawn_base_amount = ctx.accounts.credix_withdraw_request.base_amount_withdrawn;
+        std::cmp::min(
+            checked_sub(requested_base_amount, withdrawn_base_amount)?,
+            checked_sub(withdrawable_base_amount, withdrawn_base_amount)?,
+        )
```
uxd-vincent, line 355: main fix

programs/uxd/src/instructions/credix_lp/rebalance_create_withdraw_request_from_credix_lp_depository.rs
```diff
@@ -175,7 +177,7 @@ pub(crate) fn handler(
     // -- Compute the profits and overflow amounts of the depository
     // ---------------------------------------------------------------------
 
-    let redeemable_amount_under_management = checked_convert_u128_to_u64(
+    let redeemable_amount_under_management = checked_as_u64(
```
uxd-vincent, line 180: since i'm testing those features, I cleanup the maths logic on the way

programs/uxd/src/utils/math/mod.rs
```diff
@@ -1,5 +1,4 @@
 pub mod checked_add_u128_and_i128;
-pub mod checked_convert_u128_to_u64;
```
uxd-vincent, line 2: This function was created to do exactly what the maths function are doing now. Removing duplicate and using the `maths.rs` version instead

cnek (approved):

acamill (approved):
And math simplification, what else <3

PR #300, opened 2023-09-14
[Credix] Update to 0.9.1

uxd-vincent:
## TLDR Summary:
Screenshot 2023-09-15 at 12 49 33 PM

## PR Combo:
- uxd-client: https://github.com/UXDProtocol/uxd-client/pull/57
- uxd-program: https://github.com/UXDProtocol/uxd-program/pull/300
- uxd-cpi: https://github.com/UXDProtocol/uxd-cpi/pull/12

## [UXD] Changes updates recap:
 - A) We update the usage of `credix_treasury` -> `credix_treasury_pool`
 - B) We update the usage of `credix_multisig` -> `credix_treasury`
 - C) We make sure to not mistake `treasury_pool` with `credix_treasury` which are now 2 completely different things
 - D) We remove the usage of `withdraw_request`
 - E) We use the new calculation function to compute the withdrawable amount during credix redeem
 - F) We update the tests to work with latest version

### Breaking:
- feat: credix pass now has flags instead of boolean fields in the struct. All the fields are exposed by functions.
- feat: withdraw epochs now do the rebalancing, the redeem base locked for investor calculation now take request amount and lp owned both into account. (withdraw request account was removed from contexts)
- feat: Repayment schedule's periods now have calculation date. Describing on what date the accrual will be calculated from.
feat: Membership fee calculations are not based on the yearly outstanding principal
- feat: Program State now has credix treasury. Credix treasury is where the credix fees will be collected from now on.

### For Integrators:
Required changes for integrators in there smart contract:

- If any field example(is_active, is_investor) from the credix pass is used. Replace it with (is_active(), is_investor()...)
- Remove withdraw Request accounts from the cpi calls. 
- Mark withdraw epoch mutable if it is already not when creating withdraw request or redeeming withdraw requests.
- For Redeem withdraw request and withdraw funds. Change the credix_muiltisig -> credix_treasury. And Credix_muiltisig_token_account-> credix_treasury_token_account. The correct credix_treasury address is in program_state account of credix program. 

- To get the max amounts that can be withdrawn in the withdraw Epoch. We have provided a function in the rust client.

### Changes required for there local setup:
- When sending the repayment schedule with set_repayment_schedule we can directly send the RepaymentPeriod instead of RepaymentPeriodsInput struct which previously existed. The calculation_date in the periods can be set equal to the period's time_frame.start.


programs/uxd/tests/integration_tests/api/program_credix/accounts/find_pdas.rs
```diff
@@ -32,12 +32,26 @@ pub fn find_liquidity_pool_token_account(
...
+pub fn find_credix_treasury_token_account(
+    credix_treasury: &Pubkey,
+    base_token_mint: &Pubkey,
+) -> Pubkey {
+    spl_associated_token_account::get_associated_token_address(credix_treasury, base_token_mint)
+}
+
+pub fn find_treasury_pool(multisig: &Pubkey) -> Pubkey {
+    *multisig // The treasury_pool is an abritrary wallet that collect the fees
+}
```
SushantChandla, line 48: this function and `find_treasury_pool` and `find_credix_treasury` look similar.

SushantChandla (commented):
Ran tests on this branch locally by replacing the binary with the latest shared in TG. Can confirm all the test works.

(Rust Tests)

programs/uxd/tests/integration_tests/api/program_credix/accounts/find_pdas.rs
uxd-vincent: For sure, this is aimed at abstracting away how those wallets are generated in the first place, and those are two separate PDA that you mentionned have no relation to each other 👌

GregoryNEUT (approved):
Sounds good

uxd-vincent:
Gonna merge for now, so I can prepare the develop->main PR

PR #306, opened 2023-10-02
[Controller] Remove the need to register depositories

uxd-vincent:
Originally reported by @acamill , we could deprecate the use of "registered" depositories, and instead only enable mint/redeem for the router configured depositories instead, this would avoid having to "double register" both for the controller and for the router each depositories

### DIFF
--- a/programs/uxd/src/instructions/credix_lp/edit_credix_lp_depository.rs
+++ b/programs/uxd/src/instructions/credix_lp/edit_credix_lp_depository.rs
@@ -21,7 +21,6 @@ pub struct EditCredixLpDepository<'info> {
         mut,
         seeds = [CONTROLLER_NAMESPACE],
         bump = controller.load()?.bump,
-        constraint = controller.load()?.registered_credix_lp_depositories.contains(&depository.key()) @UxdError::InvalidDepository,
         has_one = authority @UxdError::InvalidAuthority,
     )]
     pub controller: AccountLoader<'info, Controller>,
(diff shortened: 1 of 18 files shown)


programs/uxd/src/instructions/credix_lp/edit_credix_lp_depository.rs
```diff
@@ -21,7 +21,6 @@ pub struct EditCredixLpDepository<'info> {
         mut,
         seeds = [CONTROLLER_NAMESPACE],
         bump = controller.load()?.bump,
-        constraint = controller.load()?.registered_credix_lp_depositories.contains(&depository.key()) @UxdError::InvalidDepository,
```
uxd-vincent, line 24: the edit IXs are permissioned, and I see no reason not to be able to edit a non-registered depository. It might be useful to be able to configure a depository before we mark it as usable by the router.

programs/uxd/src/instructions/edit_mercurial_vault_depository.rs
```diff
@@ -22,7 +22,6 @@ pub struct EditMercurialVaultDepository<'info> {
         seeds = [CONTROLLER_NAMESPACE],
         bump = controller.load()?.bump,
         has_one = authority @UxdError::InvalidAuthority,
-        constraint = controller.load()?.registered_mercurial_vault_depositories.contains(&depository.key()) @UxdError::InvalidDepository
```
uxd-vincent, line 25: same as: https://github.com/UXDProtocol/uxd-program/pull/306/files#r1342758463

acamill:
Tempting 😆

PR #309, opened 2023-10-12
[Alloyx] Followup cleanups

uxd-vincent:
Quick follow-up PR for small details after the alloyx first PR:
- Introduce a new CI job that checks that we are testing the integration tests with the latest binaries from our integration partners
- Align the alloyx-depository to be a multiple of 128 bytes


programs/uxd/tests/unit_tests/state/alloyx_vault_depository.rs
```diff
@@ -7,7 +7,7 @@ mod test_alloyx_vault_depository {
 
     #[test]
     fn test_alloyx_vault_depository_space() -> Result<()> {
-        assert_eq!(ALLOYX_VAULT_DEPOSITORY_SPACE, 1149);
+        assert_eq!(ALLOYX_VAULT_DEPOSITORY_SPACE, 1152);
```
uxd-vincent, line 10: 1024+128=1152, should be reasonably aligned?

acamill: haha yes! I don't even know the impact tbh

acamill (approved):

PR #319, opened 2023-12-05
[Alloyx] Use alloyx as emergency liquidity if identity_depository is emptied

uxd-vincent:
The TLDR of the change is to make the rebalancing flows computed by:
```rust
let deposit_redeemable_amount = std::cmp::min(
    identity_depository_overflow_value,
    alloyx_vault_depository_underflow_value,
);
let withdraw_redeemable_amount = std::cmp::max(
    identity_depository_underflow_value,
    alloyx_vault_depository_overflow_value,
);
```

This is the case so that:
 - when the identity_depository is underflowing, we withdraw even if the alloyx_depository is not overflowing
 - when the alloyx_vault_depository is underflowing, we deposit only if the identity_depository is overflowing


acamill (approved):
Looks good

programs/uxd/src/instructions/rebalance_alloyx_vault_depository.rs
```diff
@@ -141,14 +142,70 @@ pub struct RebalanceAlloyxVaultDepository<'info> {
...
+    let alloyx_vault_depository_underflow_value = alloyx_vault_depository_target_redeemable_amount
+        .saturating_sub(alloyx_vault_depository_redeemable_amount_under_management);
+
+    let identity_depository_overflow_value = identity_depository_redeemable_amount_under_management
+        .saturating_sub(identity_depository_target_redeemable_amount);
+    let alloyx_vault_depository_overflow_value =
+        alloyx_vault_depository_redeemable_amount_under_management
+            .saturating_sub(alloyx_vault_depository_target_redeemable_amount);
+
+    let deposit_redeemable_amount = std::cmp::min(
```
acamill, line 189: Given this we should be good 👍

programs/uxd/src/instructions/rebalance_alloyx_vault_depository.rs
```diff
@@ -141,14 +142,70 @@ pub struct RebalanceAlloyxVaultDepository<'info> {
...
+        identity_depository_overflow_value,
+        alloyx_vault_depository_underflow_value,
+    );
+    let withdraw_redeemable_amount = std::cmp::max(
```
acamill, line 193: If the values are equal in some scenario, what happen, does the order matter

uxd-vincent: If there is somehow a conflict, the deposited value will be used, then the withdraw value will not be used (a single rebalancing will only do a single withdraw/deposit, not both)

In all cases however, one of the two value will always be zero:
- if the overflow value > 0, then the underflow value == 0
- So there is always 2 values set to zero in the min/max expressions
- So there is always one of the two deposit/withdraw value that is zero

acamill:
Ok makes sense

PR #336, opened 2024-08-19
fix: revive testing tooling

crypto-vincent:
After some inactivity, some of the testing tooling were not working anymore.

This PR revive the whole CI pipeline to work again.

Update dependencies to:
 - Solana: `1.18.21` (same as mainnet/devnet)
 - Rust: `1.75.0` (this seems to work fine for now, but `1.80.1` doesn't work)
 - Anchor: `0.26.0` (same as anchor_gen)

Note: Credix disabled minting on devnet, so we remove the devnet+credixMint tests (those tests are still running fine in localnet)


tests/test_ci_credix_lp.ts
```diff
@@ -79,15 +78,6 @@ import {
       credixLpDepositoryEditSuite({ authority, controller });
     });
 
-    describe('credixLpDepositoryMintSuite', function () {
```
crypto-vincent, line 82: Credix disabled minting on devnet too. We just keep the test on localnet

Orelsanpls (approved):
Looks good to me

